PR #6, opened 2024-08-24
added authors

hgloeckner:
add all contributers to halodrops


hgloeckner:
> add all contributers to halodrops

I am actually not sure who should be listed as an author in the package... with this PR it would be anyone who has contributed, but maybe that's not the best solution.

PR #25, opened 2024-08-31
Calc q mixed

hgloeckner:
I looked over the q calculation and think that metpy uses the calculation over water everywhere. Since it might make more sense to calculate over mixed, I changed it to use typhon


ninarobbins (commented):
this makes sense!

hgloeckner:
I talked to Bjorn - we should use Geets function because it's reversing what the sondes do internally.

PR #28, opened 2024-09-01
Level 4

ninarobbins:
Work in progress. 
Functions to get from Level 3 to Level 4. Right now the idea is to have a class `Circle` which contains all the properties of one circle. The flight segmentation would be done before calling this class, we would get a list of all circles, then apply class Circle, then combine all the circles together into a Level_4.nc dataset.


hgloeckner:
Nice that you started this! 
Can we make this step by step? Like, first a minimal working circle class and then we add the functionality step by step?

ninarobbins:
Idea of structure now:

level3_dataset > flight segmentation gives dictionary of dataset per circle > apply all functions on class `Circle` on each circle dataset > return a new dictionary with new modified dataset per circle > concatenate all circles into one level4_dataset > write level4_dataset to netcdf file

hgloeckner:
Nice work! I think you need to add circle_fit to the environment y'all for the tests to run. 

Also, maybe we can generelize the gridded class sich that we don't need a corcle_gridded as well? At least as long as they don't do more than concat and write

ninarobbins:
> Nice work! I think you need to add circle_fit to the environment y'all for the tests to run.
> 
> Also, maybe we can generelize the gridded class sich that we don't need a corcle_gridded as well? At least as long as they don't do more than concat and write

Good idea! it should be exactly the same format, although we have to concatenate and add a "circle" dimension

pydropsonde/pipeline.py
```diff
@@ -306,13 +308,144 @@ def iterate_method_over_dataset(
...
+                if len(c["dropsondes"]["GOOD"]) >= 6
+            ]
+        )
+
+        if "HALO" in i:
+            platform_name.append("HALO")
+        elif "P3" in i:
+            platform_name.append("P3")
+        else:
+            platform_name.append("")
```
d70-t, line 358: I would strongly discourage to extract information from the file name. File names may change, or may not be present at all (e.g. for the flight segmentation from eurec4a, the recommended way to use flight segmentation information through the all_flights.yaml file which is available through a compiled release).
Instead, you could just use the platform identifier which is given within the yaml file.

```suggestion
        platform_name.append(flightinfo["platform"]
```

ninarobbins:
I want to squash this to clean up (its quite a mess), then make another pull request for each of the circle products as suggested

hgloeckner:
> I want to squash this to clean up (its quite a mess), then make another pull request for each of the circle products as suggested

Sounds good - can you make a clean branch with more than one commit on your fork that actually starts at the current main?

ninarobbins:
@hgloeckner  I'm not sure what you mean. A new branch for the changes in this pull request? I'm not sure how I would do that. My idea was to squash the commits here, since it's working. then make a new pull request for other changes. Maybe I'm misunderstanding

ninarobbins:
> @hgloeckner I'm not sure what you mean. A new branch for the changes in this pull request? I'm not sure how I would do that. My idea was to squash the commits here, since it's working. then make a new pull request for other changes. Maybe I'm misunderstanding

I'll give this cherry picking a try

ninarobbins:
Ok, then we can kinda abandon this and move the discussion to pull request #46

hgloeckner:
I think, we moved this to other PR

PR #36, opened 2024-09-03
Add vars

hgloeckner:
I added the theta_e, theta_v and T_v variable using metpy.
However, I am missing the cf-standard-name for theta_v, maybe someone can help with that.
And maybe also look if we should change a default from metpy

### DIFF
--- a/pydropsonde/helper/__init__.py
+++ b/pydropsonde/helper/__init__.py
@@ -329,7 +329,7 @@ def calc_theta_from_T(ds):
     )
     ds = ds.assign(theta=(ds.ta.dims, theta.magnitude))
     ds["theta"].attrs = dict(
-        standard_name="potential temperature",
+        standard_name="air_potential_temperature",
         long_name="potential temperature",
         units=str(theta.units),
     )
@@ -355,7 +355,7 @@ def calc_T_from_theta(ds):
     )
     ds = ds.assign(ta=(ds.ta.dims, ta.magnitude))
     ds["ta"].attrs = dict(
-        standard_name="air temperature",
+        standard_name="air_temperature",
         long_name="air temperature",
         units=str(ta.units),
     )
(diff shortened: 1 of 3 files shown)


ninarobbins:
@hgloeckner  what do you mean by change a default from metpy?

hgloeckner:
Some of the functions in metpy have default parameters and I don't know if those are appropriate for our purpose.

ninarobbins:
for the  cf-standard-name for theta_v I found virtual_potential_temperature, although not in the "official" list

ninarobbins:
> Some of the functions in metpy have default parameters and I don't know if those are appropriate for our purpose.

For the calculation of these variables, I guess this would be the ratio of the molecular weight epsilon? I think this is pretty standard and it should be okay.

d70-t:
> for the cf-standard-name for theta_v I found virtual_potential_temperature, although not in the "official" list

There's a relatively long discussion about the addition of $\Theta_v$ to the CF standard names going on at https://github.com/cf-convention/vocabularies/issues/32
It *seems* that there's finally a relatively stable agreement that the name should become `air_virtual_potential_temperature`, but it's not yet set in stone.

Generally, **only** names which are listed in the [Standard Name Table](http://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html) should be used as the `standard_name` attribute. If no fitting entry exists in the standard name table, then the `standard_name` attribute should be omitted as well, and other attributes (e.g. `long_name` and `description`) should be used to describe the variable. In addition, one may want to propose new standard names to CF conventions for later use.

As there's a currently ongoing discussion to add that variable, I wouldn't start a new one (but maybe you want to comment on that thread to show the need for that variable, and maybe to provide advice bringing it forward, I think they are still wondering on how to best describe the variable and their units).
I would **not yet** use the proposed `standard_name`, as there's still a (slight) chance of it being changed.

ninarobbins (approved):
Looks good! Runs and it makes sense to leave out the standard name for now.

hgloeckner:
should we add it then? we can always define a pipeline that does not include them for the level 3

ninarobbins:
> should we add it then? we can always define a pipeline that does not include them for the level 3

@hgloeckner I'm confused, we should leave it out for now, right?

PR #39, opened 2024-09-11
add Typhon to dependencies

ninarobbins:


hgloeckner:
This is already in the environment.yaml in main.

PR #41, opened 2024-09-13
fix neg iwv

hgloeckner:


pydropsonde/helper/__init__.py
```diff
@@ -288,7 +288,7 @@ def calc_rh_from_q(ds):
 
 
 def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt"):
-    # ds = ds.copy().sortby("alt")
+    ds = ds.copy().sortby("alt")
```
d70-t, line 291: When I see `sortby` used on datasets which I expect to be sorted already, I'm always a bit alarmed. Often, this hides errors, but produced wrong results.
Can you maybe add a comment, in which cases the dataset isn't sorted, and why sorting is always the correct way to handle the issue?
Or, if it's a hack and a good solution is complicated, add a comment indicating that this is a hack and what should be fixed when more time is available.

hgloeckner: Yes I agree.  The dataset *should* be sorted, but the wrong way around. It's descending, but the iwv calculation needs ascending (or the other way round). The catch for a non-sorted dataset happens before in the pipeline, while interpolating

hgloeckner: Maybe the more elegant way is to calculate iwv and then multiply by -1? But that seemed weird as well

d70-t: Another option might be to just reverse the order, e.g.:
```python
xr.Dataset({"a": ("b", [1,2,3])}).isel(b=slice(None, None, -1))
```

PR #42, opened 2024-09-15
adapt to cf conventions

hgloeckner:
changed iwv and theta name to cf standards

### DIFF
--- a/pydropsonde/helper/__init__.py
+++ b/pydropsonde/helper/__init__.py
@@ -46,7 +46,7 @@ l2_variables = {
         "attributes": {
             "standard_name": "relative_humidity",
             "long_name": "relative humidity",
-            "units": "",
+            "units": "1",
             "coordinates": "time lon lat alt",
         }
     },
@@ -299,11 +299,15 @@ def calc_q_from_rh_sondes(ds):
     w_s = mpcalc.mixing_ratio(e_s * units.Pa, ds.p.values * units.Pa).magnitude
     w = ds.rh.values * w_s
     q = w / (1 + w)
-    ds = ds.assign(q=(ds.rh.dims, q))
-    ds["q"].attrs = dict(
-        standard_name="specific humidity",
-        long_name="specific humidity",
-        units="kg/kg ",
+    ds = ds.assign(
+        q=(
+            ds.rh.dims,
+            q,
+            dict(
+                standard_name="specific_humidity",
+                units="1",
+            ),
+        )
     )
     return ds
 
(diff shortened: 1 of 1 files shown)


hgloeckner:
maybe someone can check if I picked the right standard name for IWV^^'

pydropsonde/helper/__init__.py
```diff
@@ -303,9 +312,18 @@ def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt"):
...
-    ds_iwv.name = "iwv"
-    ds_iwv.attrs = {"standard name": "integrated water vapor", "units": "kg/m^2"}
+
+    ds_iwv = xr.DataArray(
+        [iwv],
+        dims=[sonde_dim],
+        coords={},
+        name="iwv",
+        attrs=dict(
+            standard_name="atmosphere_mass_content_of_water_vapor",
```
tmieslinger, line 322: **TL;DR: I'd use this standard name as suggested and add a description similar to "vertically integrated water vapour between surface and aircraft altitude".**

This standard name seems to fit to the variable as far as I understand. The only thing that came to my mind on top of that is that we don't integrate over the *full* atmorspheric column as HALO is not flying at 80km altitude. I'd still use this standard name and add a description explaining that this quantity is calculated from the vertical integral from surface up to the aircraft height. 
Alternatively, we could use the more complicated standard name "mass_content_of_water_vapor_in_atmosphere_layer" and add the layer boundaries being surface and aircraft altitude. However, I think the latter is meant to describe data that have integrals over more than one layer which is what we have or want to have in the end.

tmieslinger (commented):
I'd suggest mainly to additional descriptions, one concerning the RH with respect to water (or mixed or ice?) and another description specifying the vertical integral boundaries fro IWV.

pydropsonde/helper/__init__.py
```diff
@@ -263,12 +267,17 @@ def calc_q_from_rh(ds):
...
+        q=(
+            ds.ta.dims,
+            q,
+            dict(
+                standard_name="specific_humidity",
+                units="1",
+            ),
+        )
     )
+
```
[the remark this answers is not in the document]
hgloeckner: yes and no. we calculate it only once in the pipeline currently, but there are two functions to get q from rh, because the correct calculation from the rh we get from the sondes to q is different from the reverse of "our" q-to-rh function. so I wanted to add the reverse function for that as well; just in case someone wants to reverse calculate q from rh in the interpolated dataset

pydropsonde/helper/__init__.py
hgloeckner: that's true... i would expect that above 14km not much happens regarding total water vapor content, as it is relatively cold and dry. but yes, theoretically it is not the whole water vapor content

hgloeckner:
most of this has been addressed in other PR. The iwv attributes are added in #94

PR #46, opened 2024-09-15
Level 4 basic structure and regression

ninarobbins:


ninarobbins:
I am not sure what I did here, if by cherry picking commits all the changes from previous commits were also applied up to the latest...

ninarobbins:
I think this was not the correct thing to do. So I will try this again

hgloeckner:
have you tried a `git rebase -i main`? there you can decide which commits you want to have in the branch
(you should make a copy before, otherwise the others are lost if I understand correctly)

PR #48, opened 2024-09-16
Very basic lev4

hgloeckner:
I tried to fix the level four pull requests #28 / #46 I couldn't figure out the different commits, so I made my own, but I added you as an author @ninarobbins .... I hope that's okay with you. 

However, I did change some things:
 - moved the 'get_circle_data_from_yaml' to the `processor.py` since it's not really necessary to put it into the pipeline file. 
 - added a 'get_simple_circle_yaml' to be able to process very simple intermediate yaml files and define circles just by start and end time
 - simplyfied the 'populate_circle_objects' function; that required some changes in the yaml file processing as well, but those are not too big
 - only thing: did not rename the sonde_id dimension to sounding as we discussed before that it's probably better to just keep the original ids, since they are unique. Also, I did not add any additional dimensions/variables to the circle_ds yet. I think we should talk about which dimensions should be added later; and do that shortly before combining the data again. It's fairly easy to mess the combining up and get a dataset that's 6GB big, instead of 70 MB (at least I managed to do that).
 
 I really hope the changes are acceptable for you... I think this is a sceleton where the other functions can easily be added with their own pull requests :) Also, we can discuss tomorrow. Also, I did not add something to combine the circles again, but it can be added later.

### DIFF
--- a/dropsonde.cfg
+++ b/dropsonde.cfg
@@ -4,3 +4,11 @@ data_directory = ./example_data
 [OPTIONAL]
 path_to_flight_ids = {platform}/Level_0
 path_to_l0_files = {platform}/Level_0/{flight_id}
+
+
+[processor.Gridded.add_l3_ds]
+l3_dir = ./example_data/Level_3/Level_3.nc
+
+
+[processor.Gridded.get_simple_circle_times_from_yaml]
+yaml_file = ./example_data/example_seg_simple.yml
--- a/example_data/example_seg_simple.yml
+++ b/example_data/example_seg_simple.yml
@@ -1,5 +1,6 @@
 
 flight_id: 20200122
+platform: HALO
 segments:
 - kinds:
   - circle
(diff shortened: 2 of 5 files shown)


ninarobbins:
This works well, thanks, and I agree with your changes. I will then add the functionalities each as a pull request.

hgloeckner:
> This works well, thanks, and I agree with your changes. I will then add the functionalities each as a pull request. Only super-duper minor thing is a typo in "get_simle_circle_times_from_yaml".

Haha ^^' I fixed it - if it passes everything, you can merge if you like

ninarobbins:
I think it needs some fixes based on running ruff locally? But everything passes here so I don't quite understand

ninarobbins:
@hgloeckner then I would merge this after the ruff fix pull request is in main?

hgloeckner:
yes please

Issue #50, opened 2024-09-17
Circle means for variables


hgloeckner:
I guess it would be useful to store the circle mean data for the variables in Level 4 as well. However, if we do, it would make sense to somehow weight the sondes depending on there distance to other sondes (or something similar).

Maybe we can brainstorm a bit how to best do that

Geet-George:
Hi Helene, I agree. Indeed the circle mean data are very useful. I haven't yet looked at details of how L4 is being adopted from JOANNE, but there it is already included and [the weighting is inherent in the regression](https://github.com/Geet-George/JOANNE/blob/d0d96fc29abbd4840a303a426786e1114138eb92/joanne/Level_4/rgr_fn.py#L114). So, at every altitude, we will have an oversolved system, which means that for a least-squares fit, we get an estimate of the mean and gradients along x and y, which due to how the regression works already accounts for the differences in distances between sondes.

hgloeckner:
Ah nice, 
we don't have a real level 4 yet, I just do some calculation ([code is here](https://github.com/hgloeckner/dropsondes/blob/vert_vel/droputils/circle_products.py)), but I did not do the regression yet. 

Also, in this I changed the structure a bit, because to me it does not really make sense to change the dimensions for everything; i.e. right now, level4 is just a extension of level3 with circle data added (and two new dimensions for the variables in circles being flight_id and position; which probably should be changed to circle_name). So currently there is 
ta[sonde_id, alt] as well as omega[flight_id, position, alt] in the same dataset. There are also variables added flight[sonde_id] and c_name[sonde_id], which hold the information for each sonde which flight and circle it belongs to

To me that made sense, because sonde_id is still a unique identifier and a circle should definitely be defined by the circle name and flight_id. All other information can be derived as well (and the regressed mean variables would also get dimensions flight_id and position).

ninarobbins:
@hgloeckner the way I would do it is to have only (circle, alt) as dimensions in the level 4 data, since one still has access to the level3 which contains the sonde information, and the flight segmentation can be used to find out which sondes belong to which circle. That way the level 4 contains only the circle products, and if someone wants the sonde-by-sonde information they would have a look at the level 3.

Geet-George:
> it does not really make sense to change the dimensions for everything; i.e. right now, level4 is just a extension of level3 with circle data added (and two new dimensions for the variables in circles being flight_id and position; which probably should be changed to circle_name). 

[...]

PR #54, opened 2024-09-19
Level 4 2D-fit

ninarobbins:
Addition to basic Level 4 which performs a 2d fit for a list of variables.


pydropsonde/circles.py
```diff
@@ -16,5 +21,138 @@ class Circle:
...
+        xc = [None] * len(x_coor.T)
+        yc = [None] * len(y_coor.T)
+
+        xc = np.mean(x_coor, axis=0)
+        yc = np.mean(y_coor, axis=0)
+
+        delta_x = x_coor - xc  # *111*1000 # difference of sonde long from mean long
+        delta_y = y_coor - yc  # *111*1000 # difference of sonde lat from mean lat
+
+        object.__setattr__(self, "circle_lon", circle_x)
```
hgloeckner, line 67: why are you not just adding those to an xarray? 
There is no need to store all the variables as separate attributes to the class. there can just be one xarray to which variables are added. Or is there a reason you need those separate?

pydropsonde/circles.py
```diff
@@ -16,5 +21,138 @@ class Circle:
...
+        object.__setattr__(self, "circle_lat", circle_y)
+        object.__setattr__(self, "circle_diameter", circle_diameter)
+        object.__setattr__(self, "dx", delta_x)
+        object.__setattr__(self, "dy", delta_y)
+
+        return self
+
+    def fit2d(self, var: str = None):
```
hgloeckner, line 75: when I first started the level 4 processing, tobi and I went over the fit2d and fit2d_xr function and changed some things that haven't been ideal, e.g. this u_ = u. have you looked at the [new version](https://github.com/hgloeckner/dropsondes/blob/vert_vel/droputils/circle_products.py) of the functions? But it probably goes together with the question before why you chose to add all variables as singular attributes rather than adding to an xarray.

example_data/example_seg_simple.yml
```diff
@@ -12,8 +12,3 @@ segments:
   segment_id: example_c2
   start: 2020-01-22 15:15:00
   end: 2020-01-22 15:25:00
-- kinds:
```
hgloeckner, line 15: Is there any reason to remove this? I added it, so that we can check whether we appropriately catch if there is an 'empty circle'. I'd rather include a catch for this instead of removing the example

pydropsonde/circles.py
```diff
@@ -16,5 +21,138 @@ class Circle:
...
+            if a.sum() > 4:
+                c_xc[j], c_yc[j], c_r[j], _ = cf.least_squares_circle(
+                    [
+                        (x, y)
+                        for x, y in zip(x_coor.values[:, j], y_coor.values[:, j])
+                        if ~np.isnan(x)
+                    ]
+                )
+
+        with warnings.catch_warnings():
```
hgloeckner, line 50: any reason why you catch this warning here? I probably just don't know what it's for, but usually warnings are there for a reason and suppressing them should be done carefully (but probably you know, why it occurs).

pydropsonde/circles.py
```diff
@@ -16,5 +21,138 @@ class Circle:
     segment_id: str
 
     def dummy_circle_function(self):
```
hgloeckner, line 23: once there are real functions, we can remove this. it was just to test that the iteration worked

hgloeckner (commented):
Thanks for implementing this! 
I am a bit confused about the structure and choices, but probably you can walk me through tomorrow and I'll be able to understand your thinking. 
Also, I think this really should be tested somehow. at least the fit2d function and the get_xycoordinatesj

pydropsonde/circles.py
hgloeckner: ah... it probably occurs because the example circles are too short in length... but maybe we should then just add more and catch cases where the fit does not work?

example_data/example_seg_simple.yml
ninarobbins: The code fails when there is an empty circle. Removing it is probably not the best solution, then maybe we should add a way to deal with the empty circle.

pydropsonde/circles.py
ninarobbins: I didn't know/forgot that these were changed. I'll implement the new version! I have been implementing the JOANNE functions as close as possible, but improvements are good

pydropsonde/circles.py
ninarobbins: Yes just related to the other comment, I've been simply following JOANNE

hgloeckner:
Thanks for the pull request! 

When I run this on an example folder, the resulting circle datasets has the dimensions
```
u_intercept: 10, dudx: 10, dudy: 10,
v_intercept: 10, dvdx: 10, dvdy: 10,
q_intercept: 10, dqdx: 10, dqdy: 10,
 ta_intercept: 10, dtadx: 10, dtady: 10,
p_intercept: 10, dpdx: 10, dpdy: 10)
```  

From my understanding, at this point the dimensions should only be sonde_id: 10, alt: 1460. Maybe there is something going wrong in the assignment of the variables

pydropsonde/circles.py
```diff
@@ -15,6 +19,146 @@ class Circle:
...
+        output_core_dims=["sonde_id"],
+    ):
+        """
+        Apply the fit2d method to the xarray object using xr.apply_ufunc.
+
+        :param var: Variable name for the dataset.
+        :returns: Xarray object with intercept, dudx, dudy, and original dataset.
+        """
+        # Note that we do not pass self.dx and self.dy here; they are used internally by fit2d.
+        return xr.apply_ufunc(
```
hgloeckner, line 122: connected to the dimension comment: I think the way you call xr.apply_ufunc needs to change.

(no guarantee that this works, but something like
```
def fit2d_xr(x, y, u, sonde_dim="sonde_id"):
    # input and output dims must be a list
    return xr.apply_ufunc(
        fit2d,
        x,
        y,
        u,
        input_core_dims=[[sonde_dim], [sonde_dim], [sonde_dim]],
        output_core_dims=[(), (), ()],
    )


def apply_fit2d(circle): 
    for par in tqdm(["u", "v", "q", "ta", "p"]):
        varnames = [par + "0", "d" + par + "dx", "d" + par + "dy"]

        circle = circle.assign(
            dict(
                zip(
                    varnames,
                    fit2d_xr(
                        x=circle.x,
                        y=circle.y,
                        u=circle[par],
                        sonde_dim="sonde_id",
                    ),
                )
            )
        )
    return circle
```
with calls of `get_xy_coords_for_circles` and `apply_fit2d`)

hgloeckner (changes requested):

ninarobbins:
@hgloeckner thanks for the review, I think now the dimensions should be ok!

hgloeckner:
Hello! 
Thanks for working on it - I couldn't really test it because it needs to be rebased to main. Could you do that first? 

Also, I'd still rather keep the empty-circle example and catch it instead of just removing the example. That should not be too much effort.

hgloeckner:
With an interactive rebase
``` 
git rebase -i main
```
you can also remove the commit where you removed it

pydropsonde/circles.py
```diff
@@ -15,6 +19,107 @@ class Circle:
...
+
+        delta_x = x_coor - xc  # *111*1000 # difference of sonde long from mean long
+        delta_y = y_coor - yc  # *111*1000 # difference of sonde lat from mean lat
+
+        new_vars = dict(
+            circle_flight_altitude=self.circle_ds["aircraft_geopotential_altitude"]
+            .mean()
+            .values,
+            circle_time=self.circle_ds["launch_time"].mean().values,
+            circle_lon=circle_x,
```
hgloeckner, line 67: I just realized, that circle_lon and circle_lat should probably be dependent on altitude as well? Or maybe not, I am not sure

ninarobbins: I would say they should just be a function of circle since it's only the flight track?

hgloeckner: The thing is, the sondes move. So the mean circle lat could equally be defined per altitude. I am not sure which is better

ninarobbins:
@hgloeckner if you are okay with it, maybe we could already merge so we can build on top of this?

pydropsonde/circles.py
hgloeckner: yeah, maybe it's not ^^' no idea

hgloeckner:
> @hgloeckner @tmieslinger if you are okay with it, maybe we could already merge so we can build on top of this and get closer to a working Level 4?

I am checking now. I did not know if you where completely done with it or not

pydropsonde/circles.py
```diff
@@ -15,6 +19,107 @@ class Circle:
...
+            dy=(["sonde_id", "alt"], delta_y.values),
+        )
+        self.circle_ds = self.circle_ds.assign(new_vars)
+
+        return self
+
+    @staticmethod
+    def fit2d(x, y, u, var=None):
+        # Function logic goes here
+        # No need for self here since it's now a static method
```
hgloeckner, line 80: this comment could be removed - back referencing is a bit weird; better explain in the comments/docstring, what the functions do

pydropsonde/circles.py
```diff
@@ -15,6 +19,107 @@ class Circle:
...
+        u_cal = np.where(invalid, 0, u)
+        a[invalid] = 0
+
+        a_inv = np.linalg.pinv(a)
+        intercept, dudx, dudy = np.einsum("...rm,...m->r...", a_inv, u_cal)
+
+        return intercept, dudx, dudy
+
+    def fit2d_xr(self, x, y, u, sonde_dim="sonde_id"):
+        # Apply the static method fit2d without passing self
```
hgloeckner, line 93: see above

hgloeckner (commented):
Looks nice now - only tested dx, dy though. 

The only minor things would be
- adding docstrings, describing what the functions do
- add attributes to the variables, describing them
- clean up the commit history to the essential changes (`git rebase -i`, use drop and fixup)

hgloeckner (approved):

hgloeckner:
I also realized: 
currently the [orcestra-dropsonde repo](https://github.com/orcestra-campaign/dropsondes) fails because it does not have a yaml file. it would be really great if you could add one there as well (and add the lines to the config)

ninarobbins:
> [orcestra-dropsonde repo](https://github.com/orcestra-campaign/dropsondes)

sure!

pydropsonde/circles.py
```diff
@@ -15,6 +20,179 @@ class Circle:
...
+            "name": "circle_lon",
+            "description": "Longitude of fitted circle for all regressed sondes in circle",
+            "units": str(units.degree_east),
+        }
+        circle_lat_attrs = {
+            "name": "circle_lat",
+            "description": "Latitude of fitted circle for all regressed sondes in circle",
+            "units": str(units.degree_north),
+        }
+        flight_altitude_attrs = {
```
hgloeckner, line 88: maybe we should call this circle_altitude or so? because the others are called circle_something.

hgloeckner (commented):
Thanks for adding the attributes! I haven't tested the resulting array yet, but I will soon

hgloeckner (commented):
Hello Nina, 

<s>somehow, when I run the latest state of your branch, the circle_ds in the result does not contain the circle products any more. Could you check if that's a problem in my call or a problem in your code?<s> Maybe for that it would also be good to have a write_l4 function, because right now I have to manually run the steps and that might make the results different. It would be good if `pydropsonde -c <config>` would produce a l4 that can be checked, such that we are sure the actual pipeline call works

it was a problem in my setup.

hgloeckner:
Uh but one thing that's actually important: 

$u_0$ is not the mean in the circle. You have to calculate the mean of the variables with the regression function (run_regression in [Joanne](https://github.com/Geet-George/JOANNE/blob/master/joanne/Level_4/rgr_fn.py)). I don't know if the implementation of Geet is ideal, but in principal this is what needs to be done (but that can also be another pull request).

$u_0$ is used in the regression function developed by Bjorn and Sandrine. See 5.3.2 in the [JOANNE paper](https://essd.copernicus.org/articles/13/5253/2021/essd-13-5253-2021.html)

pydropsonde/circles.py
```diff
@@ -15,6 +20,179 @@ class Circle:
...
+                self.circle_ds[varname].attrs["description"] = description
+
+                if "mean" in varname:
+                    result_units = "{:~}".format(var_units.units)
+                elif "dq" in varname:
+                    result_units = "kg / kg / m"
+                else:
+                    result_units = "{:~}".format((var_units / units.meters).units)
+
+                self.circle_ds[varname].attrs["units"] = result_units
```
hgloeckner, line 196: it would be good if the attribute assignment would happen in the `assign()` call and not have [] before the = in this code

hgloeckner (commented):
only one last minor thing :)

hgloeckner:
closes #50

ninarobbins:
@hgloeckner I made another pull request for the writing of l4, but do you think I should do it here? So far I have just been printing the dataset circle_ds at the last step in the pipeline and having a look at it after all the steps in the pipeline are executed, which worked for me and all products were there. But it's true that having a l4 dataset file would be better. 

For the kg/kg units I agree, and also for all the variables which have a standard name I would add it and use the other one as the long name. But I think I would keep the description attribute, but this we can decide based on what @d70-t and @Geet-George say.

hgloeckner:
> @hgloeckner I made another pull request for the writing of l4, but do you think I should do it here? 


Perfect :) The write should definitely be it's own pull request

pydropsonde/circles.py
```diff
@@ -164,6 +164,11 @@ def apply_fit2d(self):
                 "Zonal gradient of " + name,
                 "Meridional gradient of " + name,
             ]
+            use_names = [
+                par + "_circle_mean",
+                "derivative_of_" + par + "_wrt_x",
```
hgloeckner, line 169: currently the variable is called 'dx' so it should be "derivative_of_" + par + "_wrt_dx", but I'd actually rename dx to x. as I kind of associate dx with a difference and not with a position; but the calculations later use it as a position.

ninarobbins: good idea, I switched dx and dy to be x and y

ninarobbins:
@hgloeckner I am having issues with the poetry and toml files. I keep having conflicts with these and now when trying to solve them the checks are failing. I really don;t understand what is happening, but maybe you know?

hgloeckner:
> @hgloeckner I am having issues with the poetry and toml files. I keep having conflicts with these and now when trying to solve them the checks are failing. I really don;t understand what is happening, but maybe you know? Also I think I made a mistake in the conflict resolution, as some commits now appear twice

Hello, 
I think something is wrong with your poetry lock file (in the meantime something changed in the main branch for zarr and now you have xarray and zarr twice in your pyproject.toml, if I see it correctly). I tried to fix it but still get an error when I try running the pipeline with `pydropsonde -c ./dropsonde.cfg`: 

```

Running calculate_circle_data...
{segment_id} is an empty segment: 'lon' or 'lat' is empty.
  0%|          | 0/5 [00:00<?, ?it/s]
 40%|████      | 2/5 [00:00<00:00, 142.18it/s]
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/runner/work/pydropsonde/pydropsonde/pydropsonde/__init__.py", line 67, in main
    pi.run_pipeline(pi.pipeline, config)
  File "/home/runner/work/pydropsonde/pydropsonde/pydropsonde/pipeline.py", line 461, in run_pipeline
    previous_substep_output = run_substep(
                              ^^^^^^^^^^^^
  File "/home/runner/work/pydropsonde/pydropsonde/pydropsonde/pipeline.py", line 426, in run_substep
    output = function(previous_substep_output[intake], substep["functions"], config)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pydropsonde/pydropsonde/pydropsonde/pipeline.py", line 351, in iterate_Circle_method_over_dict_of_Circle_objects
    result = function(value, **get_args_for_function(config, function))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pydropsonde/pydropsonde/pydropsonde/circles.py", line 185, in apply_fit2d
    result_units = "{:~}".format(var_units.units)
                                 ^^^^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'units'

```

(see [here for reference](https://github.com/hgloeckner/pydropsonde/actions/runs/11685517988/job/32539142290))

hgloeckner:
Rebased and merged with #89

PR #57, opened 2024-09-21
add N variables

### LINKED ISSUE #32: Decide whether we want M and n values
The M and n in [Joanne](https://github.com/Geet-George/JOANNE/blob/master/joanne/Level_3/fn_3.py) are basically numbers that tell how many points are in each interpolation bin.  
Maybe we should decide whether we need those? (or if they are more like a meta-data-dataset)?

hgloeckner:
I added functionality to add the N values. I don't really see what good the m variables do, because people can just define it themselves from the Ns if they need this distinction in '0' '1' and '>1' values per bin. To me, it seems a bit arbitrary that we would put this way of classifying but not any other. 

closes #32

### DIFF
--- a/pydropsonde/helper/__init__.py
+++ b/pydropsonde/helper/__init__.py
@@ -299,12 +299,15 @@ def calc_q_from_rh_sondes(ds):
     w_s = mpcalc.mixing_ratio(e_s * units.Pa, ds.p.values * units.Pa).magnitude
     w = ds.rh.values * w_s
     q = w / (1 + w)
-    ds = ds.assign(q=(ds.rh.dims, q))
-    ds["q"].attrs = dict(
-        standard_name="specific humidity",
-        long_name="specific humidity",
-        units="kg/kg ",
-    )
+    try:
+        q_attrs = ds.q.attrs
+    except AttributeError:
+        q_attrs = dict(
+            standard_name="specific_humidity",
+            long_name="specific humidity",
+            units="1",
+        )
+    ds = ds.assign(q=(ds.rh.dims, q, q_attrs))
     return ds
 
 
@@ -328,12 +331,15 @@ def calc_q_from_rh(ds):
     )
 
     q = physics.vmr2specific_humidity(vmr)
-    ds = ds.assign(q=(ds.ta.dims, q))
-    ds["q"].attrs = dict(
-        standard_name="specific humidity",
-        long_name="specific humidity",
-        units="kg/kg ",
-    )
+    try:
+        q_attrs = ds.q.attrs
+    except AttributeError:
+        q_attrs = dict(
+            standard_name="specific_humidity",
+            long_name="specific humidity",
+            units="1",
+        )
+    ds = ds.assign(q=(ds.ta.dims, q, q_attrs))
     return ds
 
 
@@ -342,13 +348,21 @@ def calc_rh_from_q(ds):
     rh = physics.vmr2relative_humidity(
         vmr=vmr, p=ds.p.values, T=ds.ta.values, e_eq=physics.e_eq_mixed_mk
     )
-    ds = ds.assign(rh=(ds.q.dims, rh))
-    ds["rh"].attrs = dict(
-        standard_name="relative humidity",
-        long_name="relative humidity",
-        units="",
-        method="water until 0degC, ice below -23degC, mixed between",
-    )
+    try:
+        rh_attrs = ds.rh.attrs.update(
+            dict(
+                method="water until 0degC, ice below -23degC, mixed between",
+            )
+        )
+    except AttributeError:
+        rh_attrs = dict(
+            standard_name="relative_humidity",
+            long_name="relative humidity",
+            units="1",
+            method="water until 0degC, ice below -23degC, mixed between",
+        )
+    ds = ds.assign(rh=(ds.q.dims, rh, rh_attrs))
+
     return ds
 
 
@@ -390,12 +404,15 @@ def calc_theta_from_T(ds):
         ds.p.values * units(ds.p.attrs["units"]),
         ds.ta.values * units(ds.ta.attrs["units"]),
     )
-    ds = ds.assign(theta=(ds.ta.dims, theta.magnitude))
-    ds["theta"].attrs = dict(
-        standard_name="air_potential_temperature",
-        long_name="potential temperature",
-        units=str(theta.units),
-    )
+    try:
+        theta_attrs = ds.theta.attrs
+    except AttributeError:
+        theta_attrs = dict(
+            standard_name="air_potential_temperature",
+            long_name="potential temperature",
+            units=str(theta.units),
+        )
+    ds = ds.assign(theta=(ds.ta.dims, theta.magnitude, theta_attrs))
 
     return ds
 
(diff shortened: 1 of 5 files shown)


Geet-George:
Yeah, I can see why they seem arbitrary, but as I mentioned in #32, they are a good idea to explain the structure of lower levels of data, also keeping with CF conventions. The numbers '0', '1' and '2' are indicative of different cell methods that are added to the attributes to the variables. From the JOANNE paper, _"The variables N_p, N_ta, N_rh and N_gps provide the number of pressure, temperature, relative humidity and GPS measurements, respectively, in each altitude bin for gridding. Depending on the values of these N variables, the corresponding cell methods – denoted by the m variables – are provided. For the m variables, possible values are 0, 1 and 2 and stand for no data, interpolation and averaging, respectively."_

I agree that in most cases, the user can make a guess or have an intuition of what might have happened based on N, but it is best to be explicit about how the data has been produced. The CF conventions actually explain this very nicely in Section 7.3:

_"The default interpretation for variables that do not have the cell_methods attribute specified depends on whether the quantity is extensive (which depends on the size of the cell) or intensive (which does not). Suppose, for example, the quantities "accumulated precipitation" and "precipitation rate" each have a time axis. A variable representing accumulated precipitation is extensive in time because it depends on the length of the time interval over which it is accumulated. For correct interpretation, it therefore requires a time interval to be completely specified via a boundary variable (i.e., via a bounds attribute for the time axis). In this case the default interpretation is that the cell method is a sum over the specified time interval. This can be (optionally) indicated explicitly by setting the cell method to sum. A precipitation rate on the other hand is intensive in time and could equally well represent either an instantaneous value or a mean value over the time interval specified by the cell. In this case the default interpretation for the quantity would be "instantaneous" (which, optionally, can be indicated explicitly by setting the cell method to point). More often, however, cell values for intensive quantities are means, and this should be indicated explicitly by setting the cell method to mean and specifying the cell bounds._

_Because the default interpretation for an intensive quantity differs from that of an extensive quantity and because this distinction may not be understood by some users of the data, it is recommended that every data variable include for each of its dimensions and each of its scalar coordinate variables the cell_methods information of interest (unless this information would not be meaningful). It is especially recommended that cell_methods be explicitly specified for each spatio-temporal dimension and each spatio-temporal scalar coordinate variable."_ 

Unfortunately, for our soundings, a `cell_method` attribute will not suffice because the methods change based on the `N` for each value of `alt`.

hgloeckner:
Thanks for the explanation!

Have you somehow indicated what the 0,1 and 2 mean in Joanne? Because I think a dataset should be self-explanatory if possible.

Also, I feel interpolation is misleading for a one-value-bin. If I understand it correctly, the value is just taken with the new altitude, no matter where in the bin it was. That's not really interpolating as people expect. And if interpolation is just used in the mathematical sense of "type of estimation, a method of constructing (finding) new data points based on the range of a discrete set of known data points" (from wikipedia, sorry ^^'); then averaging is also interpolating. 

Maybe one could add an attribute like "cell_method":"mean over bin"? And then N says exactly how many values have been in the bin originally.

Geet-George:
> Have you somehow indicated what the 0,1 and 2 mean in Joanne? Because I think a dataset should be self-explanatory if possible.

Yes, the attributes for the `m_` variables explain the [flag meanings](https://github.com/Geet-George/JOANNE/blob/d0d96fc29abbd4840a303a426786e1114138eb92/joanne/Level_3/dicts.py#L213), and so they are self-explanatory.
 
> Also, I feel interpolation is misleading for a one-value-bin. If I understand it correctly, the value is just taken with the new altitude, no matter where in the bin it was. That's not really interpolating as people expect. And if interpolation is just used in the mathematical sense of "type of estimation, a method of constructing (finding) new data points based on the range of a discrete set of known data points" (from wikipedia, sorry ^^'); then averaging is also interpolating.

Ah, good point! But the interpolation (m=1) is actually for bins with no values (but still within the interpolation tolerance range. i.e. 50 m), i.e. value for that altitude is the interpolation of values from preceding and succeeding bins with values. For bins with no values and falling outside the tolerance range, there is no data (m=0). For one-value bins and multi-value bins, it is always averaging, like you rightly said. (m=2).

hgloeckner:
Ah okay ^^' thanks again. Then I have to check if I copied the interpolation correctly

Geet-George:
Yes, I'm also digging in to the code again where the `m_` vars are calculated, and I think there might be something wrong with it, because it currently does:

```python
m_p[(m_p == np.isnan)] = np.int8(0)
m_p[(m_p == 1)] = np.int8(1)
m_p[(m_p > 1)] = np.int8(2)
```
whereas I think, it should be:

```python
m_p[(m_p == np.isnan)] = np.int8(0)
m_p[(m_p == 0)] = np.int8(1)
m_p[(m_p > 0)] = np.int8(2)
```

But then again, why the `np.isnan`? The count should just be zero. I can't check currently exactly how the code runs, but it might need some bug-fixing! This is exactly where tests would have helped us four years ago. :P So, you're doing a great job including tests everywhere. :D

hgloeckner:
Not everywhere yet unfortunately ^^' I'll have a look and see.

hgloeckner:
> Yes, I'm also digging in to the code again where the `m_` vars are calculated, and I think there might be something wrong with it, because [it currently does](https://github.com/Geet-George/JOANNE/blob/d0d96fc29abbd4840a303a426786e1114138eb92/joanne/Level_3/fn_3.py#L751):
> 
> ```python
> m_p[(m_p == np.isnan)] = np.int8(0)
> m_p[(m_p == 1)] = np.int8(1)
> m_p[(m_p > 1)] = np.int8(2)
> ```
> 
> whereas I think, it should be:
> 
> ```python
> m_p[(m_p == np.isnan)] = np.int8(0)
> m_p[(m_p == 0)] = np.int8(1)
> m_p[(m_p > 0)] = np.int8(2)
> ```
> 
> But then again, why the `np.isnan`? The count should just be zero. I can't check currently exactly how the code runs, but it might need some bug-fixing! This is exactly where tests would have helped us four years ago. :P So, you're doing a great job including tests everywhere. :D

I am not sure... I think count just says nan if there is no datapoint in the dataset (in this PR I replaced that with 0). To get the m(1), I guess we need to somehow find a way to get the number of consecutive 0 values.

hgloeckner:
now I had to implement a test for the interpolation^^' 

I think it works now, but it would be great if you could have a look @Geet-George.

(with this explanation of what m should do it also makes sense to add it; I am sorry for doubting)

Geet-George:
The N and m estimation looks good to me on running the code

image

But I am not sure if the ancillary attributes made it to the dataset (see image below). Or am I looking in the wrong place?

Screenshot 2024-10-02 at 11 32 32

hgloeckner:
if I understand the [cf conventions on ancillary variables](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/build/ch03s04.html) correctly (and I am not at all sure that I do), than the ancillary keyword should go to the main variable. I.e. the ta has an attribute 'ancillary_variables', where all the describing variables are listed. 

Did I get that right? (and are those there for you?)

Geet-George:
That is also what I understood, but I don't see the attribute for `ta` in the L3 that is created. The image I pasted above is from the new L3.

hgloeckner:
You're right. the assignment wasn't wrong. it was just overwritten for ta because that's recalculated from theta later. Fixed that

Issue #58, opened 2024-09-21
update CONTRIBUTING info


tmieslinger:
the CONTRIBUTING info stil contains HALO-DROPS and also code blocks with halodrops that should be exchanged with pydropsonde.

PR #60, opened 2024-09-23
Add Level 4 products

ninarobbins:
Add the following variables to Level 4:
 
  "div",
  "vor",
  "mean_density",
  "w_vel",
  "omega"
  
  all of which have dimensions (circle, alt)

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -161,7 +161,7 @@ class Circle:
             output_core_dims=[(), (), ()],  # Output dimensions as scalars
         )
 
-    def apply_fit2d(self, alt_var="alt"):
+    def apply_fit2d(self, alt_dim="alt"):
         variables = ["u", "v", "q", "ta", "p"]
         alt_attrs = self.circle_ds[alt_var].attrs
         assign_dict = {}
@@ -217,7 +217,7 @@ class Circle:
         self.circle_ds = ds
         return self
 
-    def get_divergence(self):
+    def get_divergence(self, alt_dim="alt"):
         D = self.circle_ds.dudx + self.circle_ds.dvdy
         D_attrs = {
             "standard_name": "divergence_of_wind",
@@ -225,11 +225,11 @@ class Circle:
             "units": "m-1",
         }
         self.circle_ds = self.circle_ds.assign(
-            D=(["alt"], D.values, D_attrs)
+            D=([alt_dim], D.values, D_attrs)
         )
         return self
     
-    def get_vorticity(self):
+    def get_vorticity(self, alt_dim="alt"):
         vor = self.circle_ds.dvdx - self.circle_ds.dudy
 
         vor_attrs = {
@@ -238,11 +238,11 @@ class Circle:
             "units": "s-1",
         }
         self.circle_ds = self.circle_ds.assign(
-            vor=(["alt"], vor.values, vor_attrs)
+            vor=([alt_dim], vor.values, vor_attrs)
         )
         return self
 
-    def get_density(self, sonde_dim="sonde_id"):
+    def get_density(self, sonde_dim="sonde_id", alt_dim="alt"):
         density = hp.density(
             self.circle_ds.p,
             self.circle_ds.ta,
@@ -262,7 +262,7 @@ class Circle:
             dict(
                 density=(self.circle_ds.ta.dims, density.values, density_attrs),
                 mean_density=(
-                    ["alt"],
+                    [alt_dim],
                     density.mean(sonde_dim).values,
                     mean_density_attrs,
                 ),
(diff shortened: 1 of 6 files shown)


ninarobbins:
tests are failing due to no circle_fit module, even though I added it to the environment.yml... I don't quite get the issue, it works locally

hgloeckner:
> tests are failing due to no circle_fit module, even though I added it to the environment.yml... I don't quite get the issue, it works locally

Yes, with the change to poetry (to make this a pip package), adding dependencies to the environment.yaml is not enough. We added a section to the [Contributing section of the documentation](https://atmdrops.github.io/pydropsonde/CONTRIBUTING.html)

You have to update the pyproject.toml and the poetry.lock

pydropsonde/circles.py
```diff
@@ -15,6 +23,172 @@ class Circle:
     platform_id: str
     segment_id: str
 
-    def dummy_circle_function(self):
-        print(self.flight_id, self.segment_id)
+    def get_xy_coords_for_circles(self):
```
hgloeckner, line 26: the changes that are already part of another pull request should not be part of this one as well

pydropsonde/circles.py
```diff
@@ -15,6 +23,172 @@ class Circle:
...
+            )
+
+        return self
+
+    def get_div_and_vor(self):
+        D = self.circle_ds.dudx + self.circle_ds.dvdy
+        vor = self.circle_ds.dvdx - self.circle_ds.dudy
+
+        self.circle_ds = self.circle_ds.assign(
+            dict(div=(["alt"], D.values), vor=(["alt"], vor.values))
```
hgloeckner, line 132: maybe it would be good to add attributes here as well, explaining what the variables are and what unit they have

hgloeckner (commented):
Looks good :) except for some minor things; mainly removing the things that are already in the other pull request and cleaning up the attributes and their assignment

Geet-George:
I just remembered that there is a bug in estimating omega, so the JOANNE function needs to change a bit. See [issue here](https://github.com/Geet-George/JOANNE/issues/61) and relevant [pull request here](https://github.com/Geet-George/JOANNE/pull/63).

ninarobbins:
Thanks @hgloeckner and @Geet-George for the comments. Step by step I am implementing them :)

pydropsonde/circles.py
```diff
@@ -18,3 +24,102 @@ class Circle:
...
+    def get_vertical_velocity(self):
+        div = self.circle_ds.div.where(~np.isnan(self.circle_ds.div), drop=True).sortby(
+            "alt"
+        )
+        zero_vel = xr.DataArray(data=[0], dims="alt", coords={"alt": [0]})
+
+        height = xr.concat([zero_vel, div["alt"]], dim="alt")
+        height_diff = height.diff(dim="alt")
+
+        del_w = -div * height_diff.values
```
bpoujol, line 88: Hi all, wow everything seems to have advanced so much already ! Not sure to where leave my comment but there is a problem in this definition of vertical velocity because it assumes air is incompressible. Instead omega should be computed first by integrating divergence over pressure (which accounts for air compressibility), and then vertical velocity can be computed by dividing by rho*g. Is that clear ? I can update those two functions if you want as I have already done it for the JOANNE dataset, but I am very unfamiliar with GitHub and don't know where to do this

Geet-George: Hi @bpoujol you are the best person to help out with this indeed! :) I think the [contributing section](https://atmdrops.github.io/pydropsonde/CONTRIBUTING.html) might help you get started on how you can contribute, but if you have any questions you can always ask here.

Geet-George: Oops, just noticed some links need to be updated in there (#58). but this should be done rather quickly.

hgloeckner: Thanks so much for chimming in @bpoujol 
Geet already pointed us to you PR for JOANNE and I am sure, @ninarobbins will change this functions accordingly. 
Another way would be for you to make a PR to her repo or we could remove omega and vertical velocity from this PR and make another one for them.

bpoujol: Thank you Geet for the documentation, very useful ! For Helene and Nina : it is as you want. If Nina plans to do it then it is great ! But the function in my JOANNE PR looks quite complicated and I am sure it is possible to make it simpler.    Else, if you prefer, I can either do a PR to the repo of @ninarobbins or I wait for this PR to be incorporated in the main code before I do a second PR. Let me know what is the best !

ninarobbins: Dear @bpoujol, sorry for the super late reply, I just got back to working on this. I am happy if you do a PR to my repository, as I would prefer not to incorporate this when it is wrong. Would you still be able to do so?

bpoujol: Hi @ninarobbins , sorry for the late reply as well. I am very busy right now and probably still for the coming one or two weeks, is it fine if I only do this afterwards ?

PR #67, opened 2024-09-26
Add alt gpsalt test

hgloeckner:
We found that the surface coverage does not make too much sense in alt-space since ASPEN is shifting the whole profile downwards if there is no data close to the surface, so I changed it to live in gpsalt-space

Additionally I added a test that checks that gpsalt and alt are not too far apart (it's a bit similar, but less restrictive). 

I also added a section to the docs... In the hope that future-dropsonde people are being warned about not adding qc to the config (and also added it to the example, such that it is enabled for testing).

### DIFF
--- a/docs/source/data/qc.rst
+++ b/docs/source/data/qc.rst
@@ -1,0 +1,27 @@
+Quality Control
+---------------
+
+Between `Level_1` (product of the ASPEN software) and `Level_2` quality control checks can be applied to
+remove sondes that don't fullfill a given standard.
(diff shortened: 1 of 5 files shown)


hgloeckner:
maybe we can add this picture to the docs as well to illustrate the problem? 
lev1_both

Geet-George:
That is indeed very odd behaviour from ASPEN. Do we see this only for sondes that don't have data all the way to the surface?

PR #73, opened 2024-09-28
Global attrs

### LINKED ISSUE #23: Move global attributes to config

hgloeckner:
moved global attributes to the config file. and added the version as a standard global attribute
Close #23


Geet-George:
Excellent idea with global attrs in the config itself! How about giving global attrs a dedicated section (e.g. GLOBAL_ATTRS) and keys within the section as individual attributes, instead of bundling it in OPTIONAL? It will make the config file easier for humans to read and edit, and reduce chances of syntax errors for instance with separators.

hgloeckner:
That's such a good idea :) I changed it

Issue #64, opened 2024-09-24
too many derived variables


bjorn-stevens:
Geet and I went back and forth on this point. I'm not sure where I ended up in the end, but at the moment I find myself a skeptical of the utility of too many derived variables in the level-3 data set, especially when there is some ambiguity as to what the values mean, which is the case for most.

For instance, is IWV or theta-e the value computed from the primary (non vertically interpolated) data or the value from the interpolated data. More generally, I find the inflation of variables in the level3 data cumbersome.

I would favor a single quality assessment, i.e., good or questionable, and the amount of postprocessing should be small enough to allow this assessment. It seems that BAHAMAS data is included in this dataset, if so that doesn't seem like a good idea.
For interpolation I would interpolate  $q$ and $\theta$  regardless of what thermodynamics coordinates you decide to use in the end

I would thus prefer that level2 and level3 only differ by the use of a common grid for the latter, and that both be identified with physical dimensions as suggested in https://github.com/observingClouds/pysonde/issues/40 . Level 4 could be derived variables.

hgloeckner:
At the moment, IWV and theta_e are calculated on the interpolated grid, but that is easy to change and I wasn't sure which way is better. 

I know Geet had a similar point before... I don't really see why it is a problem to have more variables? The dataset is still very small, it's zarr and chunked, so if you don't need the variables, you don't have to look at them or make a list and just select the variables you want at the start: 
```
my_vars = ['q', 'theta']
ds = ds[my_vars]
```

Regarding ambiguity, that's a temporary state - this dataset is not yet finished and I agree, in the end the attributes to the variables have to unambigously describe what's in there. For now, we had other things that seemed more important.

There will only be Good, bad (not usable) and ugly (partly usable) in the end, but right now this is not done completely and will come alongside the flight segmentation, I guess. On the other hand, I think it is useful to know, what the quality control tests said for each sonde - also to decide if that's useful for the specific usecase at hand. 

Your wishes for physical coordinates and no-bahamas data are not feasible together. If we want to add launch-lat, launch-lon and launch-height as coordinates, we need to know where the plane has been (this is taken from the a-files of the sondes which probably ultimately is the bahamas stream; probably Geet knows that better).
Theta and q are the variables that are interpolated. RH and T are re-calculated later.

Moving derived variables to level4 is no problem; from my understanding the sole change between level3 & level4 should be the addition of circle products, but that can be changed.

bjorn-stevens:
[...]

PR #79, opened 2024-11-05
remove N and m duplicate values from dataset

hgloeckner:
remove duplicate N and m values (they are the same for theta & T as well as RH & q). 

first step to #64


hgloeckner:
(this is also already rebased to #82)

hgloeckner:
I'll change this to xhistogram on the way

PR #82, opened 2024-11-07
add circle_fit to poetry.lock

hgloeckner:

PR #84, opened 2024-11-14
Reorganize variables

hgloeckner:
already rebased to #82 and #79

this adds platform id and flight id as variables to the level 2 dataset and also makes sonde_id a coordinate. Further, it removes the theta_e, ... calculation from the standard pipeline.


hgloeckner:
With this we only have IWV, wdir, wspd, q and theta as derived variables. I think that should be fine

hgloeckner:
this is really unorganized. I'll make a new one

PR #93, opened 2024-11-25
make alt dimension change possible

hgloeckner:
due to this ASPEN processing problem (#67), we decided to make gpsalt the alititude dimension in our dataset in order to be able to also use sondes with partial data. This is the functionality to make this possible with the correct entries in the config file. 

Edit: also added a documentation entry on how to do that with the config.


ninarobbins (commented):
Nice to have this option, thank you! I tested it adding the lines to the config file, and it seems that for me "gpsalt" is not being converted into a dimension for `self._interim_l3`, which leads to an error in the function `get_chunks` in `xarray_helper`, since `chunks` has dimensions `(time, gpsalt)` and `ds` has dimensions `(time, alt)`.

hgloeckner:
Thanks! I think I fixed it. But it should be (sonde_id, gpsalt) not time in interim level 3.

PR #103, opened 2024-11-29
Empty afiles

hgloeckner:
Allow the processing of sondes that have a broken or empty afile. 

This also removes sondes that don't have any information about altitude from the pipeline between level 2.


hgloeckner:
(if you want to test this, those cases happen on HALO-20240827 and HALO-20240821)

PR #116, opened 2024-12-10
Add ORCESTRA sondes

ninarobbins:
Add sondes from ORCESTRA day to match flight segmentation. For some reason, I still do not get any segments out of the data, even though the segmentation file is properly read


example_data/HALO/Level_0/20240818/0_SysLog_20240818_093555.csv
hgloeckner, line 1: Thanks for adding those! I think it makes sense to remove all the B, C and D..._P files, as well as the aircraft data files that are not used in the pipeline

hgloeckner: (they should also be removed from previous commits. Otherwise they will still be there in the history, which takes unnecessary storage)

PR #118, opened 2024-12-11
Adjust flight id for segmentation

ninarobbins:
No segments were being read from ORCESTRA data because the flight ID from the l3_ds was 20240926 (for example), and the flight ID in the segmentation file was "HALO-20240926a". This is not a super general fix for all types of segmentation files, though, so maybe there should be an entry in the config file?

### DIFF
--- a/pydropsonde/processor.py
+++ b/pydropsonde/processor.py
@@ -1934,7 +1934,7 @@ class Gridded:
             for platform_id in platform_ids
             for flight_id in flight_ids
             for s in segmentation.get(platform_id, {})
-            .get(flight_id, {})
+            .get(f"{platform_id}-{flight_id}a", {})
             .get("segments", [])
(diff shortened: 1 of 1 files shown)


hgloeckner:
I think this has been fixed in the recent version of the pipeline. have you reprocessed to level3 (deleting the interim_l3 files first)?

ninarobbins:
I did and I still get the same discrepancy.

hgloeckner:
> I did and I still get the same discrepancy.

The folders in Level 0 should have the names of the flights. I adjusted this for IPFS, it might be that it's not working for you locally. I think we should assume that people name their folders correctly, as we cannot foresee any way that could be done different from the actual naming scheme

ninarobbins:
aha okay. So then I will change the names of the folders, makes sense. Thanks!

PR #121, opened 2024-12-12
Add orcestra data and remove extra files

ninarobbins:
Added sonde data for ORCESTRA campaign to test the real flight segmentation. Removed unneeded B, C, D_P, R, and aircraft files. 

The pipeline is failing at `create_and_populate_circle_object` for the circle on 2024-09-26


hgloeckner:
> @hgloeckner what do you think about merging this now?
> 
> Edit: I would add some more sondes to test the concatenation

I would not add too many - the idea is that the pipeline is fast and catches obvious errors. so two circles with 3 sondes each plus one that's not in a circle should be enough for testing I guess (also, just because the pipeline passes, it does not mean that everything is perfect. You should always do more elaborate testing locally and implement additional unittests).

pydropsonde/circles.py
hgloeckner, line 1: I think the commits that add the circles and pipeline file can be removed 
```
git rebase -i main
```
and then drop them

hgloeckner (commented):
Nice addition - I think some of the commits are unnecessary and can be removed or squashed together and the level 1 files from the old sondes are not needed anymore

pydropsonde/circles.py
ninarobbins: I'm not sure which ones you mean by these

hgloeckner: I think that the only commits that really belong into this PR are 
54ea068, 979a389, ec702e0

The others can be dropped because they have already been in others or should be a separate PR. 

You could also consider squashing the three I mentioned above together

PR #122, opened 2024-12-12
catch empty segments in create circles

ninarobbins:
Makes an exception when no sondes are found in a segment, because otherwise the pipeline fails when there are no sondes in a segment.

### DIFF
--- a/pydropsonde/pipeline.py
+++ b/pydropsonde/pipeline.py
@@ -249,7 +249,7 @@ def create_and_populate_flight_object(
 
 def create_and_populate_circle_object(
     gridded: Gridded, config: configparser.ConfigParser
-) -> dict:
+) -> dict[Circle]:
     """
     Create circle objects for further analysis
     """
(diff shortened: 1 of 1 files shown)


hgloeckner:
Something is really weird, this does not complain for all empty circles, just for some. 
Anyways, do something like this (it's what you are doing just not catching more errors than necessary) 

`

def create_and_populate_circle_object(
    gridded: Gridded, config: configparser.ConfigParser
) -> dict[Circle]:
    """
    create circle objects for further analysis
    """

    circles = {}

    for segment in gridded.segments:
        try:
            circle_ds = gridded.l3_ds.where(
                (gridded.l3_ds["launch_time"] > np.datetime64(segment["start"])) & (gridded.l3_ds["launch_time"] < np.datetime64(segment["end"])),
                drop=True,
            )
        except ValueError:
            print(f"No data for segment {segment["segment_id"]}")
        else:
            circle = Circle(
                circle_ds=circle_ds,
                clon=segment.get("clon"),
                clat=segment.get("clat"),
                crad=segment.get("radius"),
                flight_id=segment["flight_id"],
                platform_id=segment["platform_id"],
                segment_id=segment["segment_id"],
                alt_dim=gridded.alt_dim,
            )
            circles[segment["segment_id"]] = circle

    return circles
    
`

PR #125, opened 2024-12-13
Add omega

hgloeckner:
this contains the commits of #123 (will be removed when that is merged)

It should implement the right omega as discussed in #60 @bpoujol a plot I made looked simillar to what you did years ago for Geet, so I hope it's fine.

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -297,3 +297,32 @@ class Circle:
         }
         self.circle_ds = ds.assign(vor=(ds.dudx.dims, vor.values, vor_attrs))
         return self
+
+    def add_omega(self):
+        """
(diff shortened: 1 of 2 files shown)


pydropsonde/circles.py
bpoujol, line 1: Thank you @hgloeckner  for including this, it looks good to me ! And sorry for not having written this myself. Assuming hydrostaticity it is also possible to deduce from the pressure velocity $\omega$ the vertical velocity $w = - \omega / (\rho g)$, some people may find it convenient

hgloeckner: all good :) You wrote a pretty good explanation what should be done. 

I will add the vertical velocity as well - just wanted to be sure that everything is alright like this

PR #128, opened 2024-12-16
add wvel

hgloeckner:
add vertical velocity as calculated from omega to dataset


pydropsonde/circles.py
```diff
@@ -324,7 +324,34 @@ def add_omega(self):
...
+        """
+        ds = self.circle_ds
+        alt_dim = self.alt_dim
+        div = ds.div.where(~np.isnan(ds.div), drop=True).sortby(alt_dim)
+        zero_vel = xr.DataArray(data=[0], dims=alt_dim, coords={alt_dim: [0]})
+
+        height = xr.concat([zero_vel, div[alt_dim]], dim=alt_dim)
+        height_diff = height.diff(dim=alt_dim)
+
+        del_w = -div * height_diff.values
```
ninarobbins, line 348: Looks good! I was wondering if it's okay to assume an incompressible flow here, but I guess for the speeds we normally have it's very reasonable.

pydropsonde/circles.py
```diff
@@ -324,7 +324,34 @@ def add_omega(self):
...
+
+        w_vel = del_w.cumsum(dim=alt_dim)
+        wvel_attrs = {
+            "standard_name": "upward_air_velocity",
```
ninarobbins, line 352: "upward" means that upward is positive, right?

ninarobbins (commented):
It looks good to me!

pydropsonde/circles.py
hgloeckner: I haven't thought about it. what would be different for a compressible flow?

PR #132, opened 2025-01-17
Change good 0

hgloeckner:
Change qc flags to use 0 as the GOOD value (this should be consistent and is now similar to wales qc)


tmieslinger (approved):
Having `0 - GOOD` and leaving potential further numbers for other flags seems super reasonable to me. Thanks for adjusting the flags!
I ran the processing for the old and new flagging style and the results are as expected. Also the adjusted tests seem correct. I don't have a good overview of the whole software package at the moment and might not be able to point to a code line that should be changed but is not included in the current PR, but the suggested changes are in agreement with the intention stated :)

PR #133, opened 2025-01-17
Add profile extent qc

hgloeckner:
add a test to check that the profile extends (almost) to the aircraft

### DIFF
--- a/pydropsonde/helper/quality.py
+++ b/pydropsonde/helper/quality.py
@@ -144,7 +144,7 @@ class QualityControl:
             The sampling frequency of `time_dimension` in hertz. Default is 4.
         sparsity_threshold : float or str, optional
             The threshold for the sparsity fraction. If the calculated sparsity fraction is less than this threshold,
-            the profile is considered not full. Default is 0.8.
+            the profile is considered not full. Default is 0.2.
 
 
         """
--- a/pydropsonde/helper/quality.py
+++ b/pydropsonde/helper/quality.py
@@ -110,13 +110,13 @@ class QualityControl:
             self.qc_flags[f"{variable}_profile_extent"] = max_alt > extent_min
             self.qc_details[f"{variable}_profile_extent_max"] = max_alt
 
-    def profile_fullness(
+    def profile_sparsity(
         self,
         ds,
         variable_dict={"u": 4, "v": 4, "rh": 2, "ta": 2, "p": 2},
         time_dimension="time",
         timestamp_frequency=4,
-        fullness_threshold=0.8,
+        sparsity_threshold=0.2,
     ):
         """
         Calculates the profile coverage for a given set of variables, considering their sampling frequency.
@@ -125,11 +125,11 @@ class QualityControl:
         implying a timestamp_frequency of 4 hertz. This is applicable for ASPEN-processed QC and PQC files,
         specifically for RD41.
 
-        For each variable in the variable_dict that is in self.qc_vars, the function calculates the fullness fraction. If the fullness
-        fraction is less than the fullness_threshold, it sets the entry "profile_fullness_{variable}" in `self.qc_flag` to False.
+        For each variable in the variable_dict that is in self.qc_vars, the function calculates the sparsity fraction. If the sparsity
+        fraction is less than the sparsity_threshold, it sets the entry "profile_sparsity_{variable}" in `self.qc_flag` to False.
         Otherwise, it sets this entry to True.
 
-        For each variable in the variable_dict  that is in self.qc_vars, the function adds the fullness fraction to the qc_details dictionary
+        For each variable in the variable_dict  that is in self.qc_vars, the function adds the sparsity fraction to the qc_details dictionary
 
         Parameters
         ----------
@@ -142,8 +142,8 @@ class QualityControl:
             The independent dimension of the profile. Default is "time".
         timestamp_frequency : numeric, optional
             The sampling frequency of `time_dimension` in hertz. Default is 4.
-        fullness_threshold : float or str, optional
-            The threshold for the fullness fraction. If the calculated fullness fraction is less than this threshold,
+        sparsity_threshold : float or str, optional
+            The threshold for the sparsity fraction. If the calculated sparsity fraction is less than this threshold,
             the profile is considered not full. Default is 0.8.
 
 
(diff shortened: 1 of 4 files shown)


pydropsonde/helper/quality.py
```diff
@@ -76,6 +76,28 @@ def get_is_floater(
...
+            if no_na.size > 0:
+                max_alt = np.nanmax(no_na)
+            else:
+                max_alt = np.nan
+
+            max_diff = drop_alt - max_alt
+            if max_diff > drop_alt * float(extend_ratio):
+                self.qc_flags[f"{variable}_profile_extend"] = False
+            else:
+                self.qc_flags[f"{variable}_profile_extend"] = True
```
tmieslinger, line 98: You could remove the if statement
```suggestion
            self.qc_flags[f"{variable}_profile_extend"] = max_diff < drop_alt * float(extend_ratio)
```

pydropsonde/processor.py
```diff
@@ -464,10 +464,16 @@ def get_qc(self, run_qc=None):
...
+            run_qc = [
+                "profile_fullness",
+                "profile_extend",
+                "near_surface_coverage",
+                "alt_near_gpsalt",
+            ]
         elif isinstance(run_qc, str):
             run_qc = run_qc.split(",")
         ds = self.interim_l2_ds
+        ds.attrs.update(self.flight_attrs)
```
tmieslinger, line 476: Why do you need to update the l2_interim attrs at that point?

tmieslinger (commented):
The implementation of this new QC check seems reasonable. I'd mainly suggest to rename all "*_exten**d**" instances to "*_exten**t**" as my first intuition was that this function wants to extend the profile instead of checking the extent. 
The QC flag `profile_extend` seems to indeed happen in 12 cases in the current L3_qc data on IPFS. Once even combined with the flag `near_surface` :)
You told be before, but why exactly do we need this flag?

pydropsonde/processor.py
hgloeckner: Because in the extent_qc the aircraft altitude is needed and at this point in the pipeline that's only an attribute. maybe there is a better way to pass it on?

hgloeckner:
> The implementation of this new QC check seems reasonable. I'd mainly suggest to rename all "__exten**d**" instances to "__exten**t**" as my first intuition was that this function wants to extend the profile instead of checking the extent. The QC flag `profile_extend` seems to indeed happen in 12 cases in the current L3_qc data on IPFS. Once even combined with the flag `near_surface` :) You told be before, but why exactly do we need this flag?

I don't really understand what you mean here. the profile_extent is independent of the near_surface qc. The former checks whether the profile reaches up to the aircraft (at least sufficiently so) and the other checks that there at least 50 values close to the surface. in the `{var}_qc` variables, all qcs for that variable are combined in this byte-version, maybe that's where they appear together?

tmieslinger:
> > The implementation of this new QC check seems reasonable. I'd mainly suggest to rename all "__exten**d**" instances to "__exten**t**" as my first intuition was that this function wants to extend the profile instead of checking the extent. The QC flag `profile_extend` seems to indeed happen in 12 cases in the current L3_qc data on IPFS. Once even combined with the flag `near_surface` :) You told be before, but why exactly do we need this flag?
> 
> I don't really understand what you mean here. the profile_extent is independent of the near_surface qc. The former checks whether the profile reaches up to the aircraft (at least sufficiently so) and the other checks that there at least 50 values close to the surface. in the `{var}_qc` variables, all qcs for that variable are combined in this byte-version, maybe that's where they appear together?

Sorry, I mixed by statements :) I was wondering when or where we need to know whether the profile extends all the way to the aircraft. Is it for the integrated variables like IWV to be comparable? I am wondering because the aircraft flew at different hight levels anyway. So when is it a problem if the sonde measurements start >10% below the aircraft altitude?

hgloeckner:
> > > The implementation of this new QC check seems reasonable. I'd mainly suggest to rename all "__exten**d**" instances to "__exten**t**" as my first intuition was that this function wants to extend the profile instead of checking the extent. The QC flag `profile_extend` seems to indeed happen in 12 cases in the current L3_qc data on IPFS. Once even combined with the flag `near_surface` :) You told be before, but why exactly do we need this flag?
> > 
> > 
> > I don't really understand what you mean here. the profile_extent is independent of the near_surface qc. The former checks whether the profile reaches up to the aircraft (at least sufficiently so) and the other checks that there at least 50 values close to the surface. in the `{var}_qc` variables, all qcs for that variable are combined in this byte-version, maybe that's where they appear together?
> 
> Sorry, I mixed by statements :) I was wondering when or where we need to know whether the profile extends all the way to the aircraft. Is it for the integrated variables like IWV to be comparable? I am wondering because the aircraft flew at different hight levels anyway. So when is it a problem if the sonde measurements start >10% below the aircraft altitude?

Ah okay :) Yes, it is basically only necessary for the IWV calculation. We do have different aircraft heights, but somehow, there should still be some cutoff when it does not make sense to calculate it. I think, this test only targets one sonde that is not targeted otherwise for orcestra, which for some reason only has measurements in the last 600m. Maybe we can come up with a better way to flag sondes like that?

tmieslinger:
> > > > The implementation of this new QC check seems reasonable. I'd mainly suggest to rename all "__exten**d**" instances to "__exten**t**" as my first intuition was that this function wants to extend the profile instead of checking the extent. The QC flag `profile_extend` seems to indeed happen in 12 cases in the current L3_qc data on IPFS. Once even combined with the flag `near_surface` :) You told be before, but why exactly do we need this flag?
> > > 
> > > 
> > > I don't really understand what you mean here. the profile_extent is independent of the near_surface qc. The former checks whether the profile reaches up to the aircraft (at least sufficiently so) and the other checks that there at least 50 values close to the surface. in the `{var}_qc` variables, all qcs for that variable are combined in this byte-version, maybe that's where they appear together?
> > 
> > 
> > Sorry, I mixed by statements :) I was wondering when or where we need to know whether the profile extends all the way to the aircraft. Is it for the integrated variables like IWV to be comparable? I am wondering because the aircraft flew at different hight levels anyway. So when is it a problem if the sonde measurements start >10% below the aircraft altitude?
> 
> Ah okay :) Yes, it is basically only necessary for the IWV calculation. We do have different aircraft heights, but somehow, there should still be some cutoff when it does not make sense to calculate it. I think, this test only targets one sonde that is not targeted otherwise for orcestra, which for some reason only has measurements in the last 600m. Maybe we can come up with a better way to flag sondes like that?

Ok, when plotting IWV dependent on altitude we could define a range in altitude that adds most to the total integral. THe highest contribution should come from the BL with decreasing contributions as you go up. Do you think it could make sense to rather have a test for IWV (in the tropics) that will apply a simple fixed height threshold, e.g. the profile has to have measurements up to 10km?

tmieslinger:
From personal conversation we agreed to split the two points mentioned above:
1. providing a flag value that hints to the extent of a profile. Maybe the `profile_fullness` flag should be renamed to `profile_data_density` or similar to not confuse with `profile_extent`.
2. checking for integrated variables such as IWV that there are datapoints throughout the profile up to at least 8km (?). But the latter can be specific to the variable in question.

hgloeckner:
> From personal conversation we agreed to split the two points mentioned above:
> 
>     1. providing a flag value that hints to the extent of a profile. Maybe the `profile_fullness` flag should be renamed to `profile_data_density` or similar to not confuse with `profile_extent`.
> 
>     2. checking for integrated variables such as IWV that there are datapoints throughout the profile up to at least 8km (?). But the latter can be specific to the variable in question.

I used profile_sparsity since we calculate density later. maybe that could lead to confusion. 

Other than that, now the extent is dependent on an absolute height value again. It is set to 8000 as a default, but can be changed in the config.

tmieslinger (approved):
I tested the changes and the renaming seems to be consistent and enter the Level3 files. Should the `dropsonde.cfg` be changed accordingly to say 
```
[processor.Sonde.filter_qc_fail]
filter_flags = profile_sparcity,near_surface_coverage,alt_near_gpsalt
```

PR #135, opened 2025-01-17
Add lower physics qc

hgloeckner:
add rudimentary test for the lowest-100m-physics (currently only for rh and ta)

### DIFF
--- a/pydropsonde/helper/quality.py
+++ b/pydropsonde/helper/quality.py
@@ -222,6 +222,20 @@ class QualityControl:
             self.qc_flags["alt_near_gpsalt"] = False
         self.qc_details["alt_near_gpsalt_max_diff"] = max_diff.values
 
+    def low_physics(self, ds, alt_dim="gpsalt"):
+        ds_check = ds.where(ds[alt_dim] < 100, drop=True)
+        if ds_check.sizes["time"] == 0:
+            self.qc_flags["low_physics"] = False
+            self.qc_details["low_physics_rh_min"] = np.nan
+            self.qc_details["low_physics_ta_min"] = np.nan
+        else:
+            if (ds_check.rh.min() < 30) or (ds_check.ta.min() < 20):
+                self.qc_flags["low_physics"] = False
+            else:
+                self.qc_flags["low_physics"] = True
+            self.qc_details["low_physics_rh_min"] = ds_check.rh.min().values
+            self.qc_details["low_physics_ta_min"] = ds_check.ta.min().values
+
     def check_qc(self, used_flags=None, check_ugly=True):
         """
         check if any qc check has failed.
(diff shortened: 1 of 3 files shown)


pydropsonde/helper/quality.py
```diff
@@ -222,6 +222,37 @@ def alt_near_gpsalt(self, ds, diff_threshold=150):
...
+        """
+        ds_check = ds.where(ds[alt_dim] < 100, drop=True)
+        if ds_check.sizes["time"] == 0:
+            self.qc_flags["low_physics"] = False
+            self.qc_details["low_physics_rh_min"] = np.nan
+            self.qc_details["low_physics_ta_min"] = np.nan
+        else:
+            self.qc_flags["low_physics"] = (ds_check.rh.min() > 30) and (
+                ds_check.ta.min() > 20
+            )
```
tmieslinger, line 252: as `ta` is in kelvin, I suppose you would want to say `ds_check.ta.min() < 293.15`. 
30% RH and 20 degC is very specific for sonde profile in the tropics, while the Arctic looks very different. I'd therefore suggest to make this value changable through the config file meaning that you'd hand over an attribute in the function call, maybe with tropical defaults ;-) or designed such that the check is passed if no thresholds are defined. The latter would mean that users have to set a threshold if they want to apply this function which might be a nice hint for them.

pydropsonde/processor.py
```diff
@@ -1507,8 +1512,10 @@ def add_qc_to_interim_l3(self, keep=["sonde_qc"]):
         if keep is None:
             keep = []
         elif keep == "all":
-            keep = [f"{var}_qc" for var in list(self.qc.qc_by_var.keys())] + list(
-                self.qc.qc_details.keys()
+            keep = (
+                [f"{var}_qc" for var in list(self.qc.qc_by_var.keys())]
+                + list(self.qc.qc_details.keys())
+                + ["low_physics", "alt_near_gpsalt"]
```
tmieslinger, line 1518: Do you keep the `["low_physics", "alt_near_gpsalt"]` only and drop the `"profile_fullness", "near_surface_coverage"`? And if so, why?

tmieslinger (changes requested):
Checking the temperature and humidity in the lower altitutde seems reasonable to filter out seemingly working sondes that have unphysical values. The major change should be putting a threshold for `ta` in Kelvin.
The current Level_3 with qc flags on IPFS does not include the "low_physics" flag, is that correct? Maybe because the "keep"-statement changed?

pydropsonde/helper/quality.py
hgloeckner: yeah.... that's a problem. The conversion of celcius to kelvin happens later in the pipeline (a heritage from halodrops). I didn't want to change the whole structure for this qc. That's why it checks in Celsius. Also, RH is later transformed to "1" instead of "%".

pydropsonde/processor.py
hgloeckner: profile_fullness and near_surface coverage are variable specific, low_physics and "alt_near_gpsalt" are not (at least not really). This is why the former will be aggregated in the {var}_qc variables and the latter won't. Not adding them leads to them not being in the final dataset

pydropsonde/helper/quality.py
hgloeckner: would you change the order? Right now, qc has it's own pipeline block and the conversion to si happens in the block after (`create_L2`). We could move the `convert_to_si` at the beginning of the `qc` block.

tmieslinger: Ah, ok. In general it seems reasonable to convert to SI units as soon as possible in the pipeline if we agree that SI untis is a good thing. Would moving `convert_to_si` to teh beginning of `qc` have further impacts? I am lacking a general overview of the whole pipeline 🙈

tmieslinger (approved):
Thanks for clarifying the `ta`check!
I'd still suppose to make the threshold values changable from the config file if you also agree. Otherwise I am super happy with this PR :)

PR #137, opened 2025-01-17
Fix iwv

### LINKED ISSUE #131: remove IWV values from L3 when too few q measurements are available
In current L3 datasets, IWV is set to 0 instead of nan if no q measurements have been present. This might be missleading and probably should be set to nan.

hgloeckner:
this assures that iwv is only calculated if the rh and ta profiles are good and the surface has values.

it's marked as a draft since it relys on #136  and #135 . To fully work #133 and #132  are also needed. If those are merged, this can also go in.
Closes #131


tmieslinger (commented):
The canges in this PR that concern the IWV seem reasonable.
The commits [fix iwv check](https://github.com/atmdrops/pydropsonde/pull/137/commits/435e4a517f354283153b7d43530f912e22820ca8) and [calc iwv only for good sondes](https://github.com/atmdrops/pydropsonde/pull/137/commits/f4e91f1bd679215e4d609401d894c592b9369fc8) seem to partly overwrite each others changes. However, the overall file changes related to IWV are correct as far as I can tell.

PR #144, opened 2025-01-27
Add additional sondes from circles segmentation

hgloeckner:
add extra sondes to circle ds

add extra sondes to example data 
remove some example data that is not needed


ninarobbins (commented):
Also looks good. It makes sense to have more sondes from the segmentation and not the old day.

PR #151, opened 2025-02-06
Add circle interpolation

hgloeckner:
adds possibility to interpolate sonde profiles before calculating the circle variables
- if fewer than four values are measured by a sonde it is omitted altogether
- density is removed from the pipeline as it is not longer needed.
(rebased on PR #177  )

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -186,6 +186,39 @@ class Circle:
         )
         return self
 
+    def interpolate_na_sondes(self, method="cubic", max_gap=1500, thresh=4):
+        if method is not None:
+            ds = self.circle_ds.swap_dims({self.sonde_dim: "sonde_id"})
+            alt_dim = self.alt_dim
+            ds["p"] = np.log(ds["p"])
+            for var in [
+                var
+                for var in ds.variables
+                if set([alt_dim, "sonde_id"]).issubset(ds.variables[var].dims)
+            ]:
+                no_nan = ds[var].dropna(dim="sonde_id", thresh=int(thresh))
+                no_nan = no_nan
+                interp = no_nan.interpolate_na(
+                    dim=alt_dim,
+                    method=method,
+                    bounds_error=False,
+                    fill_value=np.nan,
+                    max_gap=int(max_gap),
+                )
+                ds = ds.assign(
+                    {
+                        var: (
+                            ds[var].dims,
+                            interp.broadcast_like(ds[var]).values,
+                            ds[var].attrs,
+                        )
+                    }
+                )
+            ds["p"] = np.exp(ds["p"])
+            self.circle_ds = ds.swap_dims({"sonde_id": self.sonde_dim})
+
+        return self
+
     @staticmethod
     def fit2d(x, y, u):
         a = np.stack([np.ones_like(x), x, y], axis=-1)
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -220,7 +220,7 @@ class Circle:
 
     def apply_fit2d(self, variables=None):
         if variables is None:
-            variables = ["u", "v", "q", "ta", "p", "density"]
+            variables = ["u", "v", "q", "ta", "p", "rh", "theta"]
         alt_var = self.alt_dim
         alt_attrs = self.circle_ds[alt_var].attrs
 
--- a/pydropsonde/pipeline.py
+++ b/pydropsonde/pipeline.py
@@ -609,7 +609,6 @@ pipeline = {
         "intake": "gridded",
         "apply": iterate_Circle_method_over_dict_of_Circle_objects,
         "functions": [
-            "add_density",
             "apply_fit2d",
             "add_divergence",
             "add_vorticity",
@@ -632,6 +631,7 @@ pipeline = {
         "intake": "gridded",
         "apply": apply_method_to_dataset,
         "functions": [
+            "drop_vars",
             "get_l4_dir",
             "get_l4_filename",
             "update_history_l4",
--- a/pydropsonde/pipeline.py
+++ b/pydropsonde/pipeline.py
@@ -601,7 +601,11 @@ pipeline = {
     "prepare_circle_dataset": {
         "intake": "gridded",
         "apply": iterate_Circle_method_over_dict_of_Circle_objects,
-        "functions": ["get_xy_coords_for_circles"],
+        "functions": [
+            "get_xy_coords_for_circles",
+            "drop_vars",
+            "interpolate_na_sondes",
+        ],
         "output": "gridded",
         "comment": "prepare circle dataset for calculation",
     },
@@ -614,7 +618,6 @@ pipeline = {
             "add_vorticity",
             "add_omega",
             "add_wvel",
-            "drop_vars",
             "add_circle_variables_to_ds",
         ],
         "output": "gridded",
@@ -631,7 +634,6 @@ pipeline = {
         "intake": "gridded",
         "apply": apply_method_to_dataset,
         "functions": [
-            "drop_vars",
             "get_l4_dir",
             "get_l4_filename",
             "update_history_l4",


pydropsonde/circles.py
```diff
@@ -147,16 +153,57 @@ def get_xy_coords_for_circles(self):
 
         return self
 
+    def apply_interp_na(self, method="cubic", max_gap=1500, x_method="cubic"):
```
tmieslinger, line 156: Some questions to the function arguments:
* what would be a use case for a different `method` compared to `x_method`?
* The `x_method` seems to be equally applied to x and y which you could communicate by calling it `xy_method`

pydropsonde/circles.py
```diff
@@ -147,16 +153,57 @@ def get_xy_coords_for_circles(self):
...
+        variables = ["u", "v", "q", "ta", "p"]
+        dss = {}
+        ds = self.circle_ds
+        alt_dim = self.alt_dim
+        ds["p"] = np.log(ds["p"])
+        if method is not None:
```
tmieslinger, line 162: It seems that you could move this check whether method is None to the very beginning of the function. At least I'd understand it such that if `method=None` `circle_ds` won't change and you skip the whole function.

pydropsonde/circles.py
```diff
@@ -147,16 +153,57 @@ def get_xy_coords_for_circles(self):
...
+                fill_value=np.nan,
+                max_gap=int(max_gap),
+            )
+            ds_y = ds.y.interpolate_na(
+                dim=alt_dim,
+                method=x_method,
+                bounds_error=False,
+                fill_value=np.nan,
+                max_gap=int(max_gap),
+            )
```
tmieslinger, line 176: Not sure whether it's elegant, but I was wondering whether you could reduce the code duplication by writing the interpolated x and y directly to the new `dss`?

```suggestion
for xy in ["x", "y"]:
    dss[xy] = ds[xy].interpolate_na(
                dim=alt_dim,
                method=x_method,
                bounds_error=False,
                fill_value=np.nan,
                max_gap=int(max_gap),
            )
```

pydropsonde/circles.py
```diff
@@ -147,16 +153,57 @@ def get_xy_coords_for_circles(self):
...
+                    dim=alt_dim,
+                    method=method,
+                    bounds_error=False,
+                    fill_value=np.nan,
+                    max_gap=int(max_gap),
+                )
+
+            ds = ds.assign({**dss, "x": ds_x, "y": ds_y})
+        ds["p"] = np.exp(ds["p"])
+        self.circle_ds = ds
```
tmieslinger, line 189: Do I understand it correctly that the variables in `circle_ds` is overwritten by the `apply_interp_na` function? (which might be aboslutely fine)

tmieslinger (commented):
Thanks for adding the possibility to interpolate the profiles along the altitude dimension!
I have a few questions, I suppose mainly for a better understanding on my side.
One general question to the PR: would it make sense to also include adding the function in the pipeline or is the function not intended to show up there?

pydropsonde/circles.py
```diff
@@ -47,6 +46,13 @@ def drop_m_N_vars(self):
 
         return self
 
+    def broadcast_ds(self):
+        circle_ds = self.circle_ds
+
+        self.circle_ds = circle_ds.where(circle_ds.u_qc)
+
+        return self
```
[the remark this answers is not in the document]
hgloeckner: There are two points to this: 
1st: It is. not needed for interpolation - therefore I removed it from this PR. 
2nd: The broadcasting is only needed if sondes (or parts of sondes) are removed; this should actually only broadcast, but there probably is a better way to do that (I will check and add it to the add_manipulating_options branch)

pydropsonde/circles.py
hgloeckner: I changed it to be only one method and further simplified it as you suggested, such that the interpolation is done on the whole dataset (for this it is necessary that there is no time variable; hence I based this on #175)

pydropsonde/circles.py
hgloeckner: ya^^' that's true

pydropsonde/circles.py
hgloeckner: as above: There is only one call now  :)

pydropsonde/circles.py
hgloeckner: Yes it is. I thought it would be less confusing. one could consider adding some qc that labels whether something is interpolated or not. On the other hand, the un-interpolated version is still in level 3 and I thought it made sense to have the data that is actually used to derive the L4 products in the L4 dataset.

pydropsonde/circles.py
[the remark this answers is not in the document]
tmieslinger: Thanks for the explanations! (also offline :))

tmieslinger (commented):
In principle I think the interpolation should work like that. However, I get an ValueError `ValueError: The number of derivatives at boundaries does not match: expected 2, got 0+0 ` which seems to come from the `cubic` interpolation method and seems to hint to the fact that there is a profile that has less than 2 values? Do you get the same error?

pydropsonde/circles.py
```diff
@@ -42,12 +41,24 @@ def drop_m_N_vars(self):
...
         return self
 
     def get_xy_coords_for_circles(self):
+        """
+        Calculate x and y from lat and lon relative to circle center.
+        """
+
         if self.circle_ds.lon.size == 0 or self.circle_ds.lat.size == 0:
             print(f"Empty segment {self.segment_id}: 'lon' or 'lat' is empty.")
             return None  # or some default value like [], np.array([]), etc.
```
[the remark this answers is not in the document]
hgloeckner: I think this comment and the one above should go into PR #177. I will adapt things there.

pydropsonde/circles.py
```diff
@@ -124,26 +149,39 @@ def get_xy_coords_for_circles(self):
...
+                ),
+                circle_lon=([], self.clon, circle_lon_attrs),
+                circle_lat=([], self.clat, circle_lat_attrs),
+                circle_radius=([], self.crad, circle_radius_attrs),
+            )
         )
+        return self
 
-        self.circle_ds = self.circle_ds.assign(new_vars)
+    def interpolate_na(self, method="cubic", max_gap=1500):
```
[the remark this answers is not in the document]
hgloeckner: Good point. I decided to rename it to `interpolate_na_sondes` because we do not interpolate anything in the circle variables. only the sondes.

tmieslinger (approved):
Juhu, the interpolation works and it's so nice to see all the gaps filled with meaningful values (I tested it on the full O data)! 🥳 
I have a tiny suggestion on the minimum # of sondes needed for interpolating a profile, but other than that, I'd merely suggest to edit the PR description a bit and maybe mention fro completeness the removal of interpolated density.

pydropsonde/circles.py
```diff
@@ -186,6 +186,39 @@ def add_circle_variables_to_ds(self):
...
+        if method is not None:
+            ds = self.circle_ds.swap_dims({self.sonde_dim: "sonde_id"})
+            alt_dim = self.alt_dim
+            ds["p"] = np.log(ds["p"])
+            for var in [
+                var
+                for var in ds.variables
+                if set([alt_dim, "sonde_id"]).issubset(ds.variables[var].dims)
+            ]:
+                no_nan = ds[var].dropna(dim="sonde_id", thresh=4)
```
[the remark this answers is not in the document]
hgloeckner: good point. I prefer making it an argument, because the threshold of 4 is only the minimum for the interpolation. One could argue, that a "profile" with only 4 points is not worth anything anyway. 
I think (but haven't checked), that this test basically deletes sondes that have no values for the given variable at all, so the number might as well be different. However, it should be adjustable.

pydropsonde/circles.py
```diff
@@ -220,7 +253,7 @@ def fit2d_xr(self, x, y, u, sonde_dim="sonde"):
 
     def apply_fit2d(self, variables=None):
         if variables is None:
-            variables = ["u", "v", "q", "ta", "p", "density"]
+            variables = ["u", "v", "q", "ta", "p", "rh", "theta"]
```
hgloeckner, line 256: maybe it's better to use the same loop as above?
```suggestion
            variables = [
                var
                for var in ds.variables
                if set([alt_dim, "sonde_id"]).issubset(ds.variables[var].dims)
            ]:
```

tmieslinger: Yes, seems consistent. I'll add it to be able to test the final code changes

PR #152, opened 2025-02-06
Weighted

hgloeckner:
based on PR #151 . 
adds the option for weighted least square fit for the circle products by either giving a path to a file with weights or a path to a file with autocorrelations. Could be expanded to other things


hgloeckner:
There is some mistake here - I will add something again once the interpolation is merged as to not make it even more confusing

PR #153, opened 2025-02-06
detect minisondes and apply minisonde profile in aspen

d70-t:
This change will automatically detect minisondes and will run the aspen CLI with the mini-dropsonde profile in these cases.

I have to admit, that I didn't see where to best put a test in the current testing framework and would be happy about ideas. My (manual) test so far was running the pipeline and checking that L1 data is generated (not 0-bytes) and that the profiles look reasonable (e.g. altitude starts at 0).

I'm not yet super happy with how `get_sonde_rev` works (it reads the `a_file` again), but I tried to stick to the surrounding style, it seems like the `a_file` is opened many times already.

I skipped defining `add_sonde_rev`, getter and `_sonde_rev` implementations, as there doesn't seem to be a technical benefit in doing so. But that disagrees with the current style. Would it be better otherwise?

### DIFF
--- a/pydropsonde/helper/rawreader.py
+++ b/pydropsonde/helper/rawreader.py
@@ -4,7 +4,7 @@ Module to read from raw files, mostly to gather metadata from A files
 
 from datetime import datetime
 import logging
-from typing import List
+from typing import List, Optional
 import os
 import fsspec
 import yaml
@@ -89,6 +89,17 @@ def get_sonde_id(a_file: "str") -> str:
     except UnboundLocalError:
         afile_base = os.path.basename(a_file)
         return afile_base.split(".")[0][1:]
+
+
+def get_sonde_rev(a_file: str) -> Optional[str]:
+    with open(a_file, "r") as f:
+        module_logger.debug(f"Opened File: {a_file=}")
+
+        for i, line in enumerate(f):
+            if "Sonde ID/Type/Rev" in line:
+                module_logger.debug(f'"Sonde ID/Type/Rev" found on line {i=}')
+                return line.split(":")[1].split(",")[2].lstrip()
+    return None
 
 
 def get_launch_time(a_file: "str") -> np.datetime64:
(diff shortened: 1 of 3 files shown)


hgloeckner:
Thank you so much for implementing this!

I think, currently there is no nice way to properly test this part of the pipeline. The only way could be to add a sonde that is known to be a minisonde to the example data, but I don't know if that's necessary; the whole testing infrastructure is not ideal. Also, since we don't really now what exactly ASPEN does it's also hard to say whether it's right. Before your implementation we got an ASPEN error (or warning) from the docker image that we don't get anymore, so it seems that at least ASPEN gets the right profile now. 

It does not seem necessary to define a "add_sonde_rev". I think most of the "add" and "get" functions come from the time where Sonde was a dataclass. It is not anymore, but the old functions haven't been changed. At some point the whole thing should be refactored properly

hgloeckner (approved):

pydropsonde/helper/rawreader.py
```diff
@@ -91,6 +91,19 @@ def get_sonde_id(a_file: "str") -> str:
...
+    with open(a_file, "r") as f:
+        module_logger.debug(f"Opened File: {a_file=}")
+        lines = f.readlines()
+
+        for i, line in enumerate(lines):
+            if "Sonde ID/Type/Rev" in line:
+                module_logger.debug(f'"Sonde ID/Type/Rev" found on line {i=}')
+                break
+
+        return lines[i].split(":")[1].split(",")[2].lstrip()
```
hgloeckner, line 104: This leads to a problem if the afile is empty (we have that for some sondes). That would lead to an UnboundLocalError. Maybe it would be better to catch that and return something as a default?

tmieslinger (approved):
Thanks for adding the minisondes!
Absolutely not relevant, bid anyone find out what "Rev" stands for? :D

pydropsonde/helper/rawreader.py
d70-t: Oh 😬 now I finally get why on earth there could be an UnboundLocalError... I'll do something about it...
Would you be fine with `None` for an unknown `sonde_rev`?

d70-t:
My guess would be **Rev**ision.

pydropsonde/helper/rawreader.py
hgloeckner: :) I think None would be a good default.

d70-t: Ok, should be fixed now.

PR #156, opened 2025-02-07
Bad gpsalt recovery

### LINKED ISSUE #69: calculate actual altitude from gpsalt and pressure
with aspen somehow moving altitude, it would be good to calculate altitude over MSL ourselves.

hgloeckner:
uses #154 

This should replace the alt dimension, if gps is not suffient and the sfc pressure is okay. I need to do some further testing, but thought it might be nice if you could already see if the general idea suits you.

Closes #69


tmieslinger (approved):
Thanks for sorting the altitude mess :)
I agree with the implemented tests and priority in in altitude sources.

PR #158, opened 2025-02-07
adapt dataset title to level

hgloeckner:
Add the level to the dataset (and the sonde_id for level2), such that not all datasets have the same title


tmieslinger (approved):
Juhuu 🎉

PR #160, opened 2025-02-12
Change l3 structure

hgloeckner:
This should address the issues mentioned in #102 for level 3.

I.e: 
- the sonde-dimension in level 3 now does not have a coordinate, it is just "sonde"
- the `launch_time` is renamed to `sonde_time` to be consistent with `circle_time` in level 4
- `sonde_id`, `platform_id` and `flight_id` are now variables as to not have string-type coordinates
- units for latitude and longitude variables are corrected
- the chunksize is adapted because chunks were super small in L3 
- when concatenating l2 drop non-matching conflicts

### DIFF
--- a/dropsonde.cfg
+++ b/dropsonde.cfg
@@ -14,9 +14,6 @@ skip=True
 [processor.Sonde.add_qc_to_interim_l3]
 keep=all
 
-[processor.Gridded.drop_qc_vars]
-keep=all
-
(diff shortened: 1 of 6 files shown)


ninarobbins (commented):
The result looks good and the structure is what we wanted also for L4, it simplifies L4 functions then.

PR #165, opened 2025-02-17
Add manipulating options

hgloeckner:
this adds manipulating options to the circles.py that can be used for testing and later for a qc measure. They are rebased on PR #151

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -219,6 +219,44 @@ class Circle:
 
         return self
 
+    def remove_values(self, n_gap=3, keep_sfc=1000):
+        n_gap = int(n_gap)
+        ds = self.circle_ds
+
+        alt_mask = np.full(ds.u.shape, False)  # first sonde_id then alt_dim
+        alt_mask[:, ::n_gap] = True
+        if keep_sfc:
+            alt_mask[:, : int(keep_sfc / 10)] = True
+        for var in ["u", "v", "rh", "q", "ta", "theta", "x", "y"]:
+            self.circle_ds = self.circle_ds.assign(
+                {var: (ds[var].dims, ds[var].where(alt_mask).values, ds[var].attrs)}
+            )
+        return self
+
+    def one_gap_one_sonde(self, alt=1500, depth=500, sonde_id=2):
+        ds = self.circle_ds
+        alt_mask = np.full(ds.u.shape, True)
+        alt_mask[
+            int(sonde_id), int(int(alt) / 10) : int((int(alt) + int(depth)) / 10)
+        ] = False
+
+        for var in ["u", "v", "rh", "q", "ta", "theta", "x", "y"]:
+            self.circle_ds = self.circle_ds.assign(
+                {var: (ds[var].dims, ds[var].where(alt_mask).values, ds[var].attrs)}
+            )
+        return self
+
+    def remove_sonde(self, sonde_id=0):
+        ds = self.circle_ds
+        alt_mask = np.full(ds.u.shape, True)
+        alt_mask[int(sonde_id), :] = False
+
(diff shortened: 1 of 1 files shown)


pydropsonde/circles.py
```diff
@@ -219,6 +219,44 @@ def interpolate_na_sondes(self, method="cubic", max_gap=1500, thresh=4):
 
         return self
 
+    def remove_values(self, n_gap=3, keep_sfc=1000):
```
d70-t, line 222: This function name is not very telling. From reading the code, I could assume that something like:
```suggestion
    def create_regular_gaps_in_data_values(self, n_gap=3, keep_sfc=1000):
```
could be more fitting.

pydropsonde/circles.py
```diff
@@ -219,6 +219,44 @@ def interpolate_na_sondes(self, method="cubic", max_gap=1500, thresh=4):
...
+        n_gap = int(n_gap)
+        ds = self.circle_ds
+
+        alt_mask = np.full(ds.u.shape, False)  # first sonde_id then alt_dim
+        alt_mask[:, ::n_gap] = True
+        if keep_sfc:
+            alt_mask[:, : int(keep_sfc / 10)] = True
+        for var in ["u", "v", "rh", "q", "ta", "theta", "x", "y"]:
```
d70-t, line 230: The division by `10` and the hardcoded list of variables seems pretty arbitrary. Thus I assume that this function was created with a specific application in mind (instead of something of generic use). But I can't find any site which uses this function, so there's also no specific application.

pydropsonde/circles.py
```diff
@@ -219,6 +219,44 @@ def interpolate_na_sondes(self, method="cubic", max_gap=1500, thresh=4):
...
+            self.circle_ds = self.circle_ds.assign(
+                {var: (ds[var].dims, ds[var].where(alt_mask).values, ds[var].attrs)}
+            )
+        return self
+
+    def one_gap_one_sonde(self, alt=1500, depth=500, sonde_id=2):
```
d70-t, line 236:
```suggestion
    def create_continuous_data_gap_in_a_single_sonde(self, alt=1500, depth=500, sonde_id=2):
```

pydropsonde/circles.py
```diff
@@ -219,6 +219,44 @@ def interpolate_na_sondes(self, method="cubic", max_gap=1500, thresh=4):
...
+            int(sonde_id), int(int(alt) / 10) : int((int(alt) + int(depth)) / 10)
+        ] = False
+
+        for var in ["u", "v", "rh", "q", "ta", "theta", "x", "y"]:
+            self.circle_ds = self.circle_ds.assign(
+                {var: (ds[var].dims, ds[var].where(alt_mask).values, ds[var].attrs)}
+            )
+        return self
+
+    def remove_sonde(self, sonde_id=0):
```
d70-t, line 249:
```suggestion
    def set_all_data_values_to_NaN(self, sonde_id=0):
```

... the function doesn't actually remove the sonde, it instead keeps the sonde and set's all their data to `NaN`.

d70-t (commented):
This PR wants to add three functions which can modify sonde data in specific, unusual ways. All functions seem to include some specific, hard-coded values, which likely renders those functions unusable in a generic case, thus it's likely that the author had some specific cases in mind. However, these specific use cases are missing in this PR and thus the functions are *unused* and therefore shouldn't be merged.

In case there's still some use for these functions, I recommend to:
* place those functions close to their specific usage site
* use more telling function names (as the functions seem to be specific, a longer name also doesn't hurt)
* use a sub-function for common code (i.e. all functions first create a mask, then apply it to the dataset. the application always is the same, so could be factored out into a separate function)

hgloeckner:
thank you for the comments. I agree - the functions are nice for playing around, but don't play any role in the pipeline. I think they should be added wherever they are used

PR #168, opened 2025-02-18
fix time encoding to resolve dates now and in the future

tmieslinger:
the previous implementation with `float32` could not resolve dates at nanosecond resolution. This PR changes it to `<i8` such that campaign data also in 100 years from now could be processed :)


hgloeckner:
Thank you very much for fixing this!

PR #173, opened 2025-02-19
change theta name to dry potential temperature

### LINKED ISSUE #164: Theta in L2
Currently, we calculate a dry-theta before the binning and interpolation.
I think that's fine, but maybe we should discuss again

hgloeckner:
Closes #164


tmieslinger (approved):
Thanks for the fix! It never hurts to be as clear a possible :)

PR #174, opened 2025-02-19
remove broken sondes from l3

### LINKED ISSUE #169: `broken_sondes` attribute
Shall we remove `broken_sondes` attribute from Level 3 and 4?
This would be in line with the argument that all QC related info is in Level2.

hgloeckner:
It feels sooo good to remove code. anyway, this will remove the `broken_sonde` attribute from L3, but keep the comment in L2, if a sonde is in the "broken_sonde" file. We could consider removing this altogether

Closes #169

### DIFF
--- a/pydropsonde/pipeline.py
+++ b/pydropsonde/pipeline.py
@@ -580,7 +580,6 @@ pipeline = {
         "functions": [
             "check_aspen_version",
             "check_pydropsonde_version",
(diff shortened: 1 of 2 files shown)


tmieslinger (approved):
I agree :) definitely removing it from Level3 after all QC has been applied.

PR #175, opened 2025-02-19
Rename and drop interp time

### LINKED ISSUE #172: rename / drop `interp_time` variable
currently, the variable `interp_time` is included in Level 3 and Level 4.
The variable name is slightly misleading and could rather be something related to `bin`, e.g. `bin_averaged_time` and a description could help, too. In Level 4 we might want to drop that variable as it's not relevant after the circle products per se and itnerested users can go back to Level3 if they need it.

hgloeckner:
- rename interp time to `bin_average_time` (#172)
- add altitude details to bin attributes
- drop `bin_average_time` and qc variables from L4

Close #172

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -25,10 +25,17 @@ class Circle:
     alt_dim: str
     sonde_dim: str
 
-    def drop_m_N_vars(self):
+    def drop_vars(self, variables=None):
         """
         drop m and N variables from level 3 from circle dataset
         """
+        if variables is None:
+            variables = [
+                "bin_average_time",
+                "alt_near_gpsalt",
+                "alt_source",
+                "alt_near_gpsalt_max_diff",
+            ]
         ds = self.circle_ds
         ds = (
             ds.drop_vars(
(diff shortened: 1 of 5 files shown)


pydropsonde/circles.py
```diff
@@ -42,6 +42,14 @@ def drop_m_N_vars(self):
             .drop_vars(
                 ["gps_m_qc", "gps_N_qc", "gpspos_N_qc", "gpspos_m_qc"], errors="ignore"
             )
+            .drop_vars(
+                [f"{var}_qc" for var in ["u", "v", "ta", "p", "rh"]],
+                errors="ignore",
+            )
```
tmieslinger, line 48: I am a fan of extending this function to drop QC and otehr vars, too :)
I was wondering whether we want to keep the `{var}_low_physics_sfc`, `{var}_p_near_surface_count`, `{var}_p_profile_extent_max`, `{var}_p_profile_sparsity_fraction` variables (in case Level 3 keeps teh QC variables) or whether we would independent of Level 3 make sure to drop those as well?

tmieslinger (approved):
Juhu! the PR adds code but cleans up the dataset :) I like it.
All changes end up as expected in the output dataset. I only have one question fro dropping even more ...

pydropsonde/circles.py
hgloeckner: that's a good point. I added the qc detail names to the ones that should be dropped. This is currently relatively messy because the circle does not inherit the qc object. but maybe it's okay for now?

tmieslinger: Yes. Something for a major refactoring at some point in the future.vThe main purpose is cleaning up the L4 dataset and that's what is done :)

tmieslinger:
Nice! No QC variables in the final L4 dataset anymore when using the default settings. And `sonde_qc` is kept 👍

PR #176, opened 2025-02-19
rearrange history and global attribute addition

hgloeckner:
This allows for level specific global attributes and structures the global attribute assignment a bit more. 

(original discussion in #70 )


pydropsonde/helper/__init__.py
```diff
@@ -154,7 +154,22 @@ def get_global_attrs_from_config(config):
...
+def get_level_specific_attrs_from_config(config):
+    try:
+        l3_attrs = dict(config.items("L3_ATTRS"))
+    except NoSectionError:
+        print("No L3 attributes in config")
+        l3_attrs = {}
+    try:
+        l2_attrs = dict(config.items("L2_ATTRS"))
+    except NoSectionError:
+        print("No L3 attributes in config")
```
tmieslinger, line 169:
```suggestion
        print("No L2 attributes in config")
```

tmieslinger (commented):
So far I found one typo ;-)
As discussed offline, I'll have a look again when L4 is there, too.

pydropsonde/helper/__init__.py
hgloeckner: ^^ I changed that part so that hopefully now there is no typo...

tmieslinger (commented):
Being able to add level specific grobal attributes is super nice!
I found one (I think) unintended behaviour. The l3 specific attributes are also part of the l4. I guess because l4 is generated out of l3 and "inherits" all attributes. Do we want to keep them as Level 4 is based on Level 3?

hgloeckner:
> Being able to add level specific grobal attributes is super nice! I found one (I think) unintended behaviour. The l3 specific attributes are also part of the l4. I guess because l4 is generated out of l3 and "inherits" all attributes. Do we want to keep them as Level 4 is based on Level 3?

That was indeed unintended. Thank you for spotting it! it would have happened between level 2 and 3 as well. I think, I fixed it now.

tmieslinger:
Great! I played around a bit and always got the expected output :)

PR #177, opened 2025-02-19
Clean get xy

hgloeckner:
- split the xy calculation from the addition of circle variables to L4
- change the circle time calculation to the mean of the first and last time.

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -274,7 +274,6 @@ class Circle:
 
             ds = self.circle_ds.assign(assign_dict)
         ds[alt_var].attrs.update(alt_attrs)
-        ds = ds.set_coords("circle_time")
 
         self.circle_ds = ds
         return self
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -73,6 +73,10 @@ class Circle:
         return self
 
     def get_xy_coords_for_circles(self):
+        """
+        Calculate x and y from lat and lon relative to circle center.
+        """
+
         if self.circle_ds.lon.size == 0 or self.circle_ds.lat.size == 0:
             print(f"Empty segment {self.segment_id}: 'lon' or 'lat' is empty.")
             return None  # or some default value like [], np.array([]), etc.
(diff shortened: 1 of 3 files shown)


pydropsonde/processor.py
```diff
@@ -1920,7 +1920,6 @@ def concat_circles(self):
         for circle_id, circle in self.circles.items():
             circle_ds = circle.circle_ds
             circle_ds = circle_ds.sortby("sonde_time")
```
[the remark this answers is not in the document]
hgloeckner: Yes... I was not sure whether it should be kept as a backup here (if segmentation is derived from something different then a flight_segmentation.yaml). but I am not sure.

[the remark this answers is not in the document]
ninarobbins: that's a good point, in the end it doesn't hurt to have it

pydropsonde/circles.py
```diff
@@ -176,7 +176,7 @@ def add_circle_variables_to_ds(self):
                 ),
                 circle_time=(
                     [],
-                    self.circle_ds["sonde_time"].mean().values,
+                    self.circle_ds["sonde_time"].isel(sonde=[0, -1]).mean().values,
```
tmieslinger, line 179: thanks for adjusting the circle "reference" time! I think that this is the better solution compared to a simple mean. Could you also adjust the `circle_time["description"]` in line 168 to reflect the change, e.g. "Mean launch time of first and last sonde in circle"?

pydropsonde/circles.py
```diff
@@ -149,27 +167,23 @@ def get_xy_coords_for_circles(self):
...
+                    circle_altitude_attrs,
+                ),
+                circle_time=(
+                    [],
+                    self.circle_ds["sonde_time"].isel(sonde=[0, -1]).mean().values,
+                    circle_time_attrs,
+                ),
+                circle_lon=([], self.clon, circle_lon_attrs),
+                circle_lat=([], self.clat, circle_lat_attrs),
+                circle_radius=([], self.crad, circle_radius_attrs),
```
tmieslinger, line 184: I am wondering whether we want to have `circle_radius`as a coordinate or "degrade" it to a variable ;-)
For circle lat and lon I can see a benefit of having those for easy plotting, but I cannot think of many applications that would need `circle_radius` as a coordinate. Any options from your side?

tmieslinger (approved):
Thanks for cleaning up this part of the code! A small step towards refactoring :)
I have two minor suggestions and apart from that the changed code runs through and gives the expected output 👍

pydropsonde/circles.py
hgloeckner: I don't have a strong opinion. and I see your point. I'll remove it.

PR #178, opened 2025-02-19
add description attribute to aircraft measurements

hgloeckner:
in reference to #64, we discussed that we would like to keep the aircraft Lon, lat and alt data to simplify plotting. However, I'd still add a hint in the attributes.

### DIFF
--- a/pydropsonde/helper/__init__.py
+++ b/pydropsonde/helper/__init__.py
@@ -115,12 +115,18 @@ l3_coords = dict(
     aircraft_longitude={
         "long_name": "aircraft longitude at launch",
         "units": "degrees_east",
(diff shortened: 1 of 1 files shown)


tmieslinger (commented):
How about calling it `source` what is named `description`?
In a description I'd expect more information content compared to the `long_name`. In the cases here, `description` provides less details than the `long_name`

hgloeckner:
> How about calling it `source` what is named `description`? In a description I'd expect more information content compared to the `long_name`. In the cases here, `description` provides less details than the `long_name`

Yes! that's way better. thank you

tmieslinger:
Tip top! The changes end up at the right place in the processed dataset :)

PR #181, opened 2025-02-21
Weighted

hgloeckner:
added the weighted functions :) 

(rebased on #165 )e

It's quite a lot... If you like I can try splitting the distance and autocorrelation calculation

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -219,30 +219,35 @@ class Circle:
         return self
 
     @staticmethod
-    def fit2d(x, y, u):
+    def fit2d(x, y, u, weight=1):
+        weight = np.asarray(weight)
         a = np.stack([np.ones_like(x), x, y], axis=-1)
 
         invalid = np.isnan(u) | np.isnan(x) | np.isnan(y)
         # remove values where fewer than 6 sondes are present. Depending on the application, this might be changed.
         under_constraint = np.sum(~invalid, axis=-1) < 6
-        u_cal = np.where(invalid, 0, u)
         a[invalid] = 0
+        sqw = np.where(invalid, 0, np.sqrt(weight))
+        wa = sqw[..., np.newaxis] * a
+        u_cal = sqw * np.where(invalid, 0, u)
 
-        a_inv = np.linalg.pinv(a)
+        a_inv = np.linalg.pinv(wa)
         intercept, dux, duy = np.einsum("...rm,...m->r...", a_inv, u_cal)
+
         intercept[under_constraint] = np.nan
         dux[under_constraint] = np.nan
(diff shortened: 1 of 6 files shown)


tmieslinger (commented):
I am having a hard time testing this PR. From going through the code I have one suggestion, but other than that, the code seems fine. 
Maybe @d70-t could have a look on this PR?

pydropsonde/processor.py
```diff
@@ -2146,6 +2147,122 @@ def add_l3_ds(self, l3_dir: str = None):
...
+        w = []
+        for var in ["u", "v", "p", "theta", "q", "rh", "ta"]:
+            if method == "autocorrelation":
+                weights = self.get_autocorrelation_weight(variable=var).transpose(
+                    self.sonde_dim, self.alt_dim
+                )
+            elif method == "no_weights":
+                weights = self.get_no_change_weights(variable=var)
+            weights = xr.where(weights < 0, 0, weights)
+            weights = xr.where(weights > 1, 1, weights)
```
d70-t, line 2256:
```suggestion
            weights = weights.clip(0, 1)
```

hgloeckner:
Thank you so much for your suggestions and comments! I think, I incorporated all of them. The main changes to the state before are: 

- correlation is calculated using SciPy.signal.correlate
- autocorrelation functions are split more and the functions to get autocorrelation and distances are outsourced to `helper.stats`
- I removed the weighting and autocorrelation part from the standard pipeline - we decided that it's a nice option for tests, but that for ORCESTRA soundings it is not necessary and hence removed. 
- the simplifications/changes for the weighted fit are also incorporated

pydropsonde/circles.py
```diff
@@ -219,33 +219,40 @@ def interpolate_na_sondes(self, method="akima", max_gap=1500, thresh=4):
...
         invalid = np.isnan(u) | np.isnan(x) | np.isnan(y)
         # remove values where fewer than 6 sondes are present. Depending on the application, this might be changed.
         under_constraint = np.sum(~invalid, axis=-1) < 6
-        u_cal = np.where(invalid, 0, u)
         a[invalid] = 0
+        sqw = np.where(invalid, 0, np.sqrt(weight))
+        wa = sqw[..., np.newaxis] * a  # np.einsum("...m,...mr->...mr", w, a)
+        u_cal = sqw * np.where(
+            invalid, 0, u
+        )  # np.einsum("...m,...m->...m", w, np.where(invalid, 0, u))
```
d70-t, line 234: I'd suggest to remove the comments and pick either implementation. (The comments don't really explain anything not written in the code, and actually seem to do something different)

d70-t (commented):
LGTM :+1:

PR #183, opened 2025-02-21
correct density calculation and fix iwv to use mixing ratio

hgloeckner:
with the switch of density in between there was a weird error in the density calculation. 
Also, I think using volume mixing ratio instead of mixing ratio was wrong. But then I am super confused as to what has to be used now. 
Maybe you can correct this mess

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -324,14 +324,14 @@ class Circle:
         ds = self.circle_ds
         assert ds.p.attrs["units"] == "Pa"
         assert ds.ta.attrs["units"] == "K"
-        density = hp.density(
+        density = hp.density_from_q(
             ds.p,
             ds.ta,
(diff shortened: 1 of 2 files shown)


tmieslinger (commented):
Thanks for solving that density-bug!
I think the [`add_density`](https://github.com/atmdrops/pydropsonde/blob/main/pydropsonde/circles.py#L314) function should be adjusted accordingly to use `density_from_q`. And maybe we can be more specific in the `long_name` that is added now that we know more about moist density :)

I tested the pipeline with a corrected `add_density` and the `density_from_mr`and `density_from_q` give the same results 👍 and the bug with the negative sign in `integrate_column` is indeed a bug.

hgloeckner:
oh true. I completely forgot about the add_density since it's currently not part of the pipeline^^' I added a commit

PR #185, opened 2025-02-25
change altitude in level 3 instead of level 2

hgloeckner:
In the previous implementation, I already removed the altitudes from level 2. That's a problem for some of the plots we want to make for the paper ^^' I think it makes sense to keep the original alt and gpsalt values in level 2 for reference

(this should only affect L2 and some qc flags and not the altitude in L3)


hgloeckner:
I also added that gpsalt is not only replaced if the near surface count in u is insufficient, but also when the extent is not reached in gpsalt, but in alt.

PR #187, opened 2025-02-27
Iterate over D-files instead of A-files

### LINKED ISSUE #167: Base pipeline iteration on D-files instead of A-files

tmieslinger:
This PR closes #167 and changes the iteration over sondes based on available A-files to D-files.
The code changes include handling the missing metadata in case an A-files does not exits. 
I am unsure whether I am missing any changes related to QC though @hgloeckner 😅 ... could you have a look there?

This PR turned into a larger list of commits due to the fact that by iterating over D-files we encountered that we have a handful of cases where two sondes have the same serial ID given by the manufacturer Vaisalia (at least that's the most likely guess at the moment). We therefore changed the processing such that it can handle duplicate serial IDs by adding a (truly) unique `sonde_id` that is a hash of the first line of a D-file. The latter includes the Vaisala serial id as well as the initialisation time stamp. We assume that this combination is unique.


hgloeckner:
The PR works for me - it finds 1177 sondes (instead of 1188 in the current main), but I think that the 11 missing might only have an a-file with no data attached. 

However, in the current main 1115 sondes make it to Level 3 and only 1114 in this branch. I think that has nothing to do with this PR though. While testing I realised that the missing sonde id 000007500 appears (at least) twice in the raw data: In the A20240924_191601.4 and A20240924184709.4 (and the corresponding d-files). The first has not detected a launch and the second has; somehow we were lucky in main that it used the launch detect-file and in this branch it uses the other a-file. That's definitely a problem, but I think it's independent of this branch.

pydropsonde/helper/paths.py
```diff
@@ -128,6 +128,14 @@ def get_all_afiles(self):
         self.afiles_list = a_files
         return a_files
 
+    def get_all_dfiles(self):
+        """Returns a list of paths to all D-files for the given directory
+        and also sets it as attribute named 'dfiles_list'
+        """
+        d_files = glob.glob(os.path.join(self.l0_dir, "D????????_??????.?"))
```
hgloeckner, line 135: There are some Dfiles of the form D??????.? ( in HALO-20240827a and HALO-20240821a) Some of them are not empty, so maybe we should at least try to run ASPEN on them as well?

tmieslinger: a regex is used to find both D-file name variants now

hgloeckner: That's nice, I think it finds all files now, but I get an `UnboundLocalError: cannot access local variable 'line_id' where it is not associated with a value ` error if I try to run it on the complete Orcestra data. I think that happens if the a-file exists but is empty. It is raised in the `check_launch_detect_in_afile` in the helper.paths. I think it happens because the loop does not find a line that contains the "Launch Obs Done?" so it won't set `line_id`. Maybe that should be caught.

hgloeckner: I think it should be `None`, because if there is no a-file, we only don't know if a launch was detected. It somehow seems connected to the gps-receiver, but I have no idea why that should be the case. Sondes with a `False` launch detect are directly discarded, which we don't want to do at this point for empty a-file sondes

tmieslinger: I agree to have `True`, `False`, and `None` as valid values for `launch_detect` and will implement your suggestions accordingly.

tmieslinger: I think I fixed it now. The pipeline works for empty and non-existing A-files

pydropsonde/helper/paths.py
```diff
@@ -153,21 +165,21 @@ def populate_sonde_instances(self, config) -> Dict:
...
 
         Sondes = {}
 
-        for a_file in afiles:
-            sonde_id = rr.get_sonde_id(a_file)
-            try:
+        for d_file in dfiles:
+            sonde_id = rr.get_sonde_id(d_file)
+            if pp(d_file.replace("D", "A")).is_file():
+                a_file = d_file.replace("D", "A")
```
d70-t, line 175: This is actually a good use case for the walrus operator:
```suggestion
            if pp(a_file := d_file.replace("D", "A")).is_file():
```

pydropsonde/helper/rawreader.py
```diff
@@ -40,69 +40,61 @@ def check_launch_detect_in_afile(a_file: "str") -> bool:
...
-                break
-
-        return bool(int(lines[line_id].split("=")[1]))
+            for i, line in enumerate(lines):
+                if "Launch Obs Done?" in line:
+                    line_id = i
+                    module_logger.debug(f'"Launch Obs Done?" found on line {line_id=}')
+                    break
 
+            return bool(int(lines[line_id].split("=")[1]))
```
d70-t, line 54: This still uses `line_id` from inside the for loop, which might lead to an unbound local error. Instead, it probably would be more elegant to directly return from within the loop an use `line` instead of `lines[line_id]`.

pydropsonde/helper/rawreader.py
```diff
@@ -40,69 +40,61 @@ def check_launch_detect_in_afile(a_file: "str") -> bool:
...
-
-            return lines[i].split(":")[1].split(",")[0].lstrip()
+        with open(d_file, "r") as f:
+            module_logger.debug(f"Opened File: {d_file=}")
+            header = f.readline()
+            return header.split(" ")[2]
     except UnboundLocalError:
-        afile_base = os.path.basename(a_file)
-        return afile_base.split(".")[0][1:]
+        dfile_base = os.path.basename(d_file)
```
d70-t, line 80: There's probably no chance anymore to end up with an unbound local here anymore.

pydropsonde/helper/rawreader.py
```diff
@@ -40,69 +40,61 @@ def check_launch_detect_in_afile(a_file: "str") -> bool:
...
+        return dfile_base.split(".")[0][1:]
 
 
-def get_sonde_rev(a_file: str) -> Optional[str]:
-    with open(a_file, "r") as f:
-        module_logger.debug(f"Opened File: {a_file=}")
+def get_sonde_rev(a_file: str | None) -> Optional[str]:
+    if a_file is not None and os.path.getsize(a_file) > 0:
```
d70-t, line 85: This likely works as intended, but one could think of using a guard clause in order to reduce the indentation depth afterwards. E.g. if the function would start with:

```python
if a_file is None:
    return None
```

The rest of the function could be as before, and with one indentation level less.

pydropsonde/helper/rawreader.py
tmieslinger: I changed it to `IndexError` as it could potentially happen that there is a D-file with a different header and less items after splitting the line

d70-t: Apparently, there's still an `UnboundLocalError`, maybe there's a commit missing?

pydropsonde/processor.py
```diff
@@ -659,23 +652,28 @@ def get_flight_attributes(
...
+                        break
+
+                attr = l2_flight_attributes_map.get(attr, attr)
+                try:
+                    value = lines[line_id].split("= ")[1]
+                    flight_attrs[attr] = float(value) if "AVAPS" not in attr else value
+                except UnboundLocalError:
+                    print(
+                        f"No flight attributes for sonde {self.serial_id} on {self.flight_id}"
+                    )
```
d70-t, line 675: This would also be a good case for a guard clause (which also would shorten the diff a bit):
```suggestion
        if not self.afile:
            print(
                f"No flight attributes for sonde {self.serial_id} on {self.flight_id}"
            )
            return self
        with open(self.afile, "r") as f:
            lines = f.readlines()

        for attr in l2_flight_attributes_map.keys():
            for line_id, line in enumerate(lines):
                if attr in line:
                    break

            attr = l2_flight_attributes_map.get(attr, attr)
            try:
                value = lines[line_id].split("= ")[1]
                flight_attrs[attr] = float(value) if "AVAPS" not in attr else value
            except UnboundLocalError:
                print(
                    f"No flight attributes for sonde {self.serial_id} on {self.flight_id}"
                )
```

pydropsonde/processor.py
```diff
@@ -659,23 +652,28 @@ def get_flight_attributes(
...
+            for attr in l2_flight_attributes_map.keys():
+                for line_id, line in enumerate(lines):
+                    if attr in line:
+                        break
+
+                attr = l2_flight_attributes_map.get(attr, attr)
+                try:
+                    value = lines[line_id].split("= ")[1]
+                    flight_attrs[attr] = float(value) if "AVAPS" not in attr else value
+                except UnboundLocalError:
```
d70-t, line 672: there's still an `UnboundLocalError`, because of the use of `line_id` from the inside of the for loop.

pydropsonde/processor.py
```diff
@@ -385,12 +378,12 @@ def filter_no_launch_detect(self) -> None:
             If the `launch_detect` attribute does not exist.
         """
         if hasattr(self, "launch_detect"):
-            if not self.launch_detect:
+            if self.launch_detect is True or self.launch_detect is None:
+                return self
+            else:
```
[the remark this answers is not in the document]
tmieslinger: I tried to be verbose there by explicitly stating that it can be `True` or `None`

pydropsonde/helper/rawreader.py
tmieslinger: Sorry, I didn't push the changes as I wanted to understand why we are lacking a sonde. Anyway, the commits are there now.

PR #190, opened 2025-02-27
change interpolation method to akima

hgloeckner:
It seems like akima is more robust to outliers and has less overshoot. That helps the interpolation in edge cases.


tmieslinger (approved):
I agree! From the general behaviour of cubic versus Akima-spline interpolation and based on plots showing the difference in profiles I think that changing the default to "Akima" is a good idea.

PR #201, opened 2025-03-14
extrapolate_sfc.

hgloeckner:
This PR should extrapolate the surface for 300m before the circle fit is done. The solution is not super ideal as it has to be applied after the other interpolation to not interpolate any gaps smaller than 300m. This comes at the benefit of being easy code.

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -227,7 +227,7 @@ class Circle:
             {
                 var: (
                     ds[var].dims,
-                    ds[var].bfill(dim="altitude", limit=int(max_alt // 10)).values,
+                    ds[var].bfill(dim=self.alt_dim, limit=int(max_alt // 10)).values,
                     ds[var].attrs,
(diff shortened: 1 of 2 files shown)


ninarobbins (commented):
Looks good. This is interpolating all gaps up to 300m, even if they are not at the surface. I guess that's still okay for IWV. Do we want to keep the original profiles in the final dataset and only use the interpolation for IWV? Or in the end have the interpolated profiles in the dataset?

hgloeckner:
> Looks good. This is interpolating all gaps up to 300m, even if they are not at the surface. I guess that's still okay for IWV. Do we want to keep the original profiles in the final dataset and only use the interpolation for IWV? Or in the end have the interpolated profiles in the dataset?

This is only for the circles. the iwv interpolation happens before that. I also changed it to only interpolate, u, v and p and to be more robust if it is done before the rest of the interpolation.

PR #203, opened 2025-03-15
extrapolate surface before l3 iwv calc

ninarobbins:
It's the same function as for the circle datasets. I'm not sure if the variables should stay interpolated for the rest of the pipeline or just for the IWV calculation.


hgloeckner:
I guess this one can be closed now as there is a new PR #214

PR #211, opened 2025-03-24
add QR code generator for pydropsonde

tmieslinger:
I thought of adding a script that generates a QR code pointing to `pydropsonde`.
At this stage, only the default works well, the optional logo embedded as image in the QR code does not work yet as I couldn't find an elegant way to convert `svg` images to a format that pillow can handle (currently it's assumed that imagemagick is installed on your machine) and also, the current default logo is too large and covers the full QR code.
If someone has an idea, let me know :)

Maybe it would be even nice to include such a script in the workflows and point to e.g. the latest release?

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -14,8 +14,10 @@ dependencies:
   - netcdf4
   - numpy
   - openssl
+  - pillow
   - pre-commit
   - pytest
+  - qrcode
   - requests
   - sphinx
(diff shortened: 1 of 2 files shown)


tmieslinger:
@d70-t do you have an idea on the svg handling? Or maybe the whole idea of an automatically generated QR code doesn't make sense and should be dropped entirely?

PR #213, opened 2025-06-12
Rename vars

hgloeckner:
This pull requests solves multiple minor requests:

- add d{var}dx and d{var}dy for theta, ta, q, rh, p to the dataset
- add a passed p_qc as a requirement to the iwv calculation, since p is also used
- rename variables such that they are closer to rapsody, i.e. 
            - w_dir and w_spd to wdir and wspd
            - sonde_time to launch_time

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -294,8 +294,8 @@ class Circle:
                 "p",
                 "rh",
                 "theta",
-                "w_dir",
-                "w_spd",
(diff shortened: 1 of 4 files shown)


tmieslinger (approved):
Thanks for adding those changes!
I ran the code on the example data and it produces the described changes meaning that everything ends up in the Level3 and Level 4 datasets as expected or intended :)

PR #214, opened 2025-07-08
Interpolate from surface before calculating IWV

ninarobbins:
The variables q, p, and T are filled in from the surface if the first 300m are empty:
- q: last valid value is extrapolated to the surface.
- p: follows hydrostatic balance.
- ta: follows a linear increase towards the surface based on a lapse rate.

I also had to adjust the interpolation function because i kept getting negative IWV.

### DIFF
--- a/pydropsonde/helper/__init__.py
+++ b/pydropsonde/helper/__init__.py
@@ -323,52 +323,54 @@ def calc_rh_from_q(ds, alt_dim="altitude"):
     return ds
 
 
-def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt", max_alt=300, qc_var=None):
+def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt", max_gap=300, qc_var=None):
     """
     Input :
 
         ds : Dataset
         sonde_dim : Dimension name for the sonde identifier
         alt_dim : Dimension name for the altitude
-        max_alt : Maximum altitude to consider for integrated water vapor calculation
+        max_gap : Maximum one-sided gap at the surface to fill for IWV calculation (m)
         qc_var : List of quality control variable names to check for valid data
 
     Output :
 
         dataset : Dataset with integrated water vapor
 
-    Function to estimate integrated water vapor in the given dataset.
+    Function to estimate integrated water vapor in the given dataset. q, p and ta are interpolated before the calculation, and up to max_gap m are extrapolated to the surface. If the gap at the surface is larger than 300m or one of the QC was not passed, Nan is returned.
     """
     if qc_var is not None:
         qc_vals = [ds[var].values for var in qc_var]
     if (qc_var is None) or (qc_vals.count(0) == len(qc_vals)):
-        alt = ds[alt_dim]
-        pressure = ds.p
-        temperature = ds.ta
-        q = ds.q
-
-        q_interp = q.interpolate_na(dim=alt_dim, method="linear")
-        log_p = np.log(pressure)
+        q_interp = ds.q.interpolate_na(dim=alt_dim, method="linear").interpolate_na(
+            dim=alt_dim, method="nearest", fill_value="extrapolate", max_gap=max_gap
+        )
+        log_p = np.log(ds.p)
         p_interp = np.exp(
-            log_p.interpolate_na(dim=alt_dim, method="linear", fill_value="extrapolate")
+            log_p.interpolate_na(dim=alt_dim, method="linear").interpolate_na(
+                dim=alt_dim, method="linear", fill_value="extrapolate", max_gap=max_gap
+            )
         )
-        ta_interp = temperature.interpolate_na(
-            dim=alt_dim, method="linear", fill_value="extrapolate"
+        ta_interp = ds.ta.interpolate_na(dim=alt_dim, method="linear").interpolate_na(
+            dim=alt_dim, method="linear", fill_value="extrapolate", max_gap=max_gap
         )
 
-        q_surface_filled = q_interp.bfill(dim=alt_dim, limit=int(max_alt / 10))
-        if np.isnan(q_surface_filled.sel(altitude=0)):
-            return np.nan
+        if (
+            (np.isnan(q_interp.sel(altitude=0)))
+            or (np.isnan(p_interp.sel(altitude=0)))
+            or (np.isnan(ta_interp.sel(altitude=0)))
+        ):
+            iwv = np.nan
         else:
             mask_p = ~np.isnan(p_interp)
             mask_t = ~np.isnan(ta_interp)
-            mask_q = ~np.isnan(q_surface_filled)
+            mask_q = ~np.isnan(q_interp)
             mask = mask_p & mask_t & mask_q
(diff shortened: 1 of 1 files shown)


pydropsonde/helper/physics.py
```diff
@@ -85,24 +85,21 @@ def integrate_water_vapor(p, q, T=None, z=None, axis=0):
             q: specific humidity
             T: temperature
             z: height
-
     """
 
     def integrate_column(y, x, axis=0):
-        if np.all(x[:-1] >= x[1:]):
+        if T is None and z is None and np.all(x[:-1] >= x[1:]):
```
hgloeckner, line 91: it is never a good idea to just change a function without understanding why it gives a different result. In this case, you change the arguments that you pass to the integrate_column function in the calc_iwv function: before it was `da.values` and now it is only `da`. This is why the order of x does not matter anymore. The way you change the integrate column function solves your problem, but it potentially breaks the calculation if someone gives np arrays (that are ordered in the opposite direction). 

I think it is fine to change that in the integrate_column function, but then you should get rid of the if statement altogether and write a doctoring saying that it needs ordered arrays or an xr.DataArray. (and check if that works of course).

pydropsonde/helper/__init__.py
```diff
@@ -338,21 +357,59 @@ def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt", qc_var=None):
...
+        g = constants.gravity_earth
+        Rd = constants.Rd
+
+        lapse_rate = (
+            (ds.ta.isel(**{alt_dim: idx_ta + 1}) - val_ta)
+            / (ds[alt_dim].isel(**{alt_dim: idx_ta + 1}) - alt_ta)
+        ).where(mask_ta)
+
+        q_surface_filled = xr.where((alt < alt_q) & mask_q, val_q, q)
+        p_surface_filled = xr.where(
```
hgloeckner, line 392: A bit similar to the ta - I guess (but you might check that) that it is enough to linearly interpolate in log-p from the measurements we have (at least that's what we do for level 4). If we only do it for 300m anyways that should be fine - and you could save the whole problem of getting the lowest valid value.

hgloeckner (changes requested):
Thank you for making this PR. 

I would suggest to make the extrapolation easier and maybe check whether the interpolation is really necessary or if we can save that. If you disagree - please fell free to discuss!

Also, your first commit (7ae869cb6797e3ea9b0d1bc5988872a84c6ed7a2) contains changes that have nothing to do with the iwv calculation, so please remove that commit. You undo those changes with the later commits anyways, so there is no reason to have them in there at all.

pydropsonde/helper/__init__.py
```diff
@@ -338,21 +357,59 @@ def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt", qc_var=None):
...
+        val_q = val_q.where(mask_q)
+        val_p = val_p.where(mask_p)
+        val_ta = val_ta.where(mask_ta)
+
+        dz = alt_p - alt
+
+        g = constants.gravity_earth
+        Rd = constants.Rd
+
+        lapse_rate = (
```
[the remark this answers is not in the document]
ninarobbins: Actually, i dont extrapolate q above the last value, so then the contribution from the high levels even if i extrapolate T is nan into the IWV.

pydropsonde/helper/__init__.py
```diff
@@ -323,11 +323,28 @@ def calc_rh_from_q(ds, alt_dim="altitude"):
     return ds
 
 
-def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt", qc_var=None):
+def get_first_valid_value_from_surface(var_ds, alt_dim="alt"):
```
hgloeckner, line 326: I think this function is not needed anymore and can be removed when the extrapolation is done differently

pydropsonde/helper/__init__.py
```diff
@@ -338,21 +355,36 @@ def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt", qc_var=None):
...
+        q_interp = q.interpolate_na(dim=alt_dim, method="linear")
+        log_p = np.log(pressure)
+        p_interp = np.exp(
+            log_p.interpolate_na(dim=alt_dim, method="linear", fill_value="extrapolate")
+        )
+        ta_interp = temperature.interpolate_na(
+            dim=alt_dim, method="linear", fill_value="extrapolate"
+        )
+
+        q_surface_filled = q_interp.bfill(dim=alt_dim, limit=int(max_alt / 10))
```
hgloeckner, line 372: it might be worth checking if after this the lowest value is still nan and if so return nan for the iwv. Just in case at a later time there are sondes with a surface gap of more than 300m

pydropsonde/processor.py
```diff
@@ -1442,6 +1442,72 @@ def interpolate_alt(
         self.interim_l3_ds = interp_ds
         return self
 
+    def extrapolate_na_sondes(self, max_alt=300):
```
hgloeckner, line 1445: This function should not be in this PR

hgloeckner (changes requested):
Thank you for the adjustments! The iwv calculation works for me and it seems cleaner to me.

However, I would still ask you to clean the history in this branch such that there are not things added in the first commit and then later removed again. Other than that, I think it is ready.

hgloeckner (commented):
Hello, 

I rebased your PR and squashed it a bit. Could you review again if everything is okay now?

hgloeckner (approved):

tmieslinger (approved):
Thank you both for improving the IWV interpolation!!
Overall, this PR changes the IWV calculation by first interpolating temperature, pressure and specific humidity including an extrapolation to the surface for T and p. For q the extrapolation to the surface is limited to a defined maximal gap (default 300m), otherwise the IWV results in NaN. I think this is a very reasonable change and I processed the full ORCESTRA data with the new code and looked at the changes which seem to be limited to very few sondes with significant improvement in IWV due to large gaps in q near the surface and negligible changes for the other sondes. Overall, the results seem correct and I only have two very minor suggestions for code changes. 
This PR should be merged soon and definitely included in the next dataset release :)

pydropsonde/helper/__init__.py
```diff
@@ -338,21 +344,59 @@ def calc_iwv(ds, sonde_dim="sonde_id", alt_dim="alt", qc_var=None):
...
-        q = ds.q.values
-        alt = ds[alt_dim].values
-
-        mask_p = ~np.isnan(pressure)
-        mask_t = ~np.isnan(temperature)
-        mask_q = ~np.isnan(q)
+        alt = ds[alt_dim]
+        pressure = ds.p
+        temperature = ds.ta
+        q = ds.q
```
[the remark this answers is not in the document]
hgloeckner: Thanks for pointing that out! I also realised that it is a little cleaner to also extrapolate p and ta only up to max_gap meters, not only q.  We also realized during the GATE comparison that bfill always fills max_gap values, even if the gap is larger. although it wouldn't break this PR, I changed it to a nearest-neighbour extrapolation.

PR #217, opened 2025-08-07
change reading code for a-file to handle P3 cases

tmieslinger:
When reprocessing the EUREC4A P3 files we encountered decoding problems of some (6) A-files. Those files include bytes that are not included in ASCII. However, such files should only include characters that are part of the ASCII table. Problematic bytes occur at typical locations which leads to the assumption that they are likely due to data transmission problems of the measurement system (very old hardware likely using RS232). As there is no way to repair the original files, I checked the location of the problematic bytes in relation to the information that pydropsonde reads from A-files and changed the reading routine for A-files to:
* open the A-file in binary mode
* search for a specific character string (with ASCII encoding)
* read in only the relevant part, either up to the end of line(default) or up to a specific `end_char` which can be any character.

In summary, the same input is read into pydropsonde from the A-files. The reading and especially decoding is narrowed down to the few relevant bytes that we are interested in such that the rest can potentially be buggy.


hgloeckner (approved):
Thank you very much for implementing this! I implemented a small test to understand what's going on and pushed it - feel free to remove it if you don't like it.

Then I have a small comment:
- the name of `find_line_in_afile` is a bit confusing, because it does not always find a complete line, nor is it limited to one line. Maybe we could write a telling dock string or rename to something like `find_string_in_afile`. 
The second point is a bit more general - please feel free to completely dismiss it (if you want it, I can make the changes as I already have them locally).
- while testing I thought it would be nicer to return the results of the `find_line_in_afile` as a list instead of a long string. (one would only have to change the return to `return res.split(end_char)` and the try block to `res = content[pos+len(search_phrase):end_pos].decode("ascii").strip()`
- if that change is implemented, the other functions of course have to be changed a bit, but only because the split does not have to happen there
- also, I'd then give the whole string as a search (eg. `"Launch Obs Done? (0,1) ="`, if that's not effected by the problems in the Eureca data)

PR #228, opened 2025-09-09
L3 bin2int

hgloeckner:
This PR adds a valid option to interpolate to the bin center in level 3. In order to do this, values with the same gpsalt values in level 2 have to be dropped. There are very few sondes that have those (I added one to the example data). 

An alternative could be to average measurements from the same altitude. Maybe we could make an issue? for now I think the loss is very limited.


hgloeckner:
sooo... there was a bug in the removal of non-monotonically increasing altitude values. That should be fixed now, along with a properly working interpolation method. 
I also found two lines in the code where I did not properly rename (they could get their own PR if you like :)

PR #233, opened 2025-09-26
Remove duplicate l2

### LINKED ISSUE #225: duplicate sonde and serial ID entries (variable and attribute) in Level 2
In our L2 datasets we seem to save the `vaisala_serial_id` and `sonde_id` as variables as well as attributes (here named `sonde_serial_ID` and `sonde_ID`). It might be slightly confusing to find the same information at two different locations and under slightly varying names. Also, `platform_id` is given as variable and attribute. `launch_time` on the other hand is provided only as attribute, while I would have expected to find it in a variable. Does anyone remember whether there were good reasons to those choices?

hgloeckner:
This closes #225. 

It 
- adds launch_time as a variable to L2
- removes platform_id, sonde_id,  serial_id and launch_time from attributes, as they are also variables

### DIFF
--- a/pydropsonde/pipeline.py
+++ b/pydropsonde/pipeline.py
@@ -549,6 +549,7 @@ pipeline = {
             "add_sonde_id_variable",
             "add_platform_id_variable",
             "add_flight_id_variable",
+            "add_launch_time_variable",
             "add_qc_to_l2",
             "get_l2_filename",
             "update_history_l2",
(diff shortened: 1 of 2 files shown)


tmieslinger (approved):
I agree with your changes and would prefer the variables just the way you implemented it now. I checked that the duplicates are gone and launch time is indeed the first time value in the time variable. Thanks for sorting this out!

PR #234, opened 2025-09-26
Std names

hgloeckner:
I hope this makes the dataset a bit cleaner. 
it 
- uses the same notation for the unit K in theta as in T
- changes the unit for q to kg/kg
- fixes the standard names of the derivatives and vorticity (I hope)
- removes the standard name for circle means, as there seems to be none suitable
- adds standard name "deployment_*" for launch lat and Lon.

### DIFF
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -585,12 +585,10 @@ class Circle:
                         )
                     )
                     .broadcast_like(ds[div_error_name])
-                    .values
-                    * 0.01
-                    * 60**2,
+                    .values,
                     dict(
                         standard_name=f"{omega_std_name} standard_error",
-                        units=ds["omega"].attrs.get("units", "hPa hr-1"),
+                        units=ds["omega"].attrs.get("units", "Pa s-1"),
                     ),
                 )
             }
@@ -752,13 +750,13 @@ class Circle:
         if p.sizes["altitude"] > 0:
             pres_diff = xr.concat([zero_vel, p.diff(dim=alt_dim)], dim=alt_dim)
             del_omega = -div * pres_diff.values
-            omega = del_omega.cumsum(dim=alt_dim) * 0.01 * 60**2
+            omega = del_omega.cumsum(dim=alt_dim)
         else:
             omega = xr.DataArray(data=[np.nan], dims=alt_dim, coords={alt_dim: [0]})
         omega_attrs = {
             "standard_name": "vertical_air_velocity_expressed_as_tendency_of_pressure",
             "long_name": "Area-averaged atmospheric pressure velocity (omega)",
-            "units": "hPa hr-1",
+            "units": "Pa s-1",
         }
         self.circle_ds = ds.assign(
             dict(omega=(ds.div.dims, omega.broadcast_like(ds.div).values, omega_attrs))
--- a/pydropsonde/circles.py
+++ b/pydropsonde/circles.py
@@ -609,7 +609,13 @@ class Circle:
             errors="ignore",
         )
         if "iwv_mean" in self.circle_ds.variables:
-            self.circle_ds["iwv_mean"] = self.circle_ds["iwv_mean"].mean("altitude")
+            self.circle_ds = self.circle_ds.assign(
+                iwv_mean=(
+                    "circle",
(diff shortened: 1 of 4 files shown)


tmieslinger (commented):
The intended changes are all there!
I am still unsure about the standard names for derivatives. One minor issue is that they include a white space presumably instead of an underscore `_ `? And in general they are not listed in the standard name table of the cf conventions. As I am unsure of the final conclusion in offline discussions, I'll simply list all cases that could be of interest in my list of potential additional and similar changes below. Another general question would be whether we want to use the units stated in the cf convention table if we also use the respective standard name? 

Here is the list level-by-level with some changes translating from one to the other:
Level 2:
* `lat` should have unit `degree_north` instead of `degrees_north`
* `lon` respectively the  unit `degree_east`
*  `rh` should have unit `1` following the cf conventions
* Standard names for `var_qc`for var in {u, v} are `status_flag`. Same as in mentioned below for Level 3, I was wondering whether the standard name `aggregate_quality_flag` (`quality_flag`?) would be more adequate? 
* The variables `* _profile_extent_max` have the unit of the respective variable. I would have expected this to be in `m`?
* `time` could get the additional info 'time_zone': 'UTC'

Level 3:
*  `rh` should have unit `1` following the cf conventions
* `lat` and `launch_lat ` should have unit `degree_north` instead of `degrees_north`
* `lon` and `launch_lon` respectively the  unit `degree_east`
* Standard names for `qc for var` with var element of {p, rh, ta, u, v} is set to `status_flag`. I am wondering whether the standard name `aggregate_quality_flag` would be more adequate? It’s also used for the variable `sonde_qc` which is in terms of the aggregation a level higher I’d say. But also the `qc for var` aggregates several flags. Alternatively, the latter could also have the standard name `quality_flag`?
* The `interpolated_time` could get the additional info 'time_zone': 'UTC' as the `launch_time` variable also includes the time zone

Level 4:
* `circle_lat` and `launch_lat ` unit `degree_north` instead of `degrees_north`
* `circle_lon` and `launch_lon` respectively the  unit `degree_east`
* `var_dpdx` and  `var_dpdy` for var in {p, q, rh, ta, theta, u, v} and all their related variables with the name extension `_std_error`, as well as `div_std_error`, `omega_std_error`, `vor_std_error` and `wvel_std_error` have non-standard names that should be translated into a long_name or description
* `iwv_mean` is lacking any attributes. Maybe we can add {'long_name': 'circle mean of integrated water vapor, 'units': 'kg m-2’}?
* `omega` currently has the correct standard name, but the unit 'hPa hr-1' while it should be 'Pa s-1’ according to cf conventions shall we convert that before writing it to the output file? If yes, the variable `omega_std_error` should be changed accordingly
*  `rh` and `rh_mean` should have unit `1` following the cf conventions. I suppose this happens automatically if Level 3 is changed.

General for all `altitude` variables in Level 3 and 4 I was wondering whether we should add somehow that it’s above mean sea level (MSL)?

hgloeckner:
Uh, thank you for the detailed review and checking the data so thoroughly. I'll briefly answer to some things 

- in Section 4.1. Latitude Coordinate of the CF convention it says "The recommended value of the units attribute is the string degrees_north" (likewise for longitudes) that's why that's our default. The other is accepted as well - I think the important point is to be consistent between datasets 
- the standard error variables in level 4 have the standard error addition to the standard name as in 3.4. Ancillary Data example 3.3; so if I understand correctly, there should be a white space
- with the derivatives, I really don't know. I think for the winds they are in the table and for some of the others they are not (or at least there is an explanation on how to do it). I don't have strong opinions, whether to have standard names for those. Maybe it's cleaner to leave them?

hgloeckner:
Ah, and the hPa hr-1 was a request by Bjorn at some point. But I think you are right and we should do that calculation for the plots and not for the data

tmieslinger:
* Ah you are so right with the lat/lon units! I was misleaded by the standard name table.
* and also on the standard name pattern for standard_error. 
* For derivatives, we could stay with the correct pattern though it's not (yet) in the official table for all variables? One small change however could be the renaming to `eastward_derivative_of_var` and northward_derivative_of_var` as those are explicitly listed in the table. It's mentioned that `x`and `y` instead of eastward and northward are ok, too, but in our case it's indeed the latter and therefore maybe better to use those terms?

